PR #3, opened 2015-03-02
Add Coinprism to the URL whitelist

Flavien:
The coinprism.com domain name needs to be whitelisted, otherwise the TREZOR won't accept calls coming from Coinprism.


prusnak:
Thanks, will release soon!

PR #9, opened 2016-03-07
Add Quroum Wallet to the URL whitelist

axic:
Quorum is an Ethereum wallet, which supports message signing features of Trezor to authorize transactions. See more on [Reddit](https://www.reddit.com/r/ethereum/comments/498ltg/quorum_using_bitcoin_hardware_wallets_to_secure/) or in [this blog post](https://medium.com/@alexberegszaszi/quorum-using-bitcoin-keys-with-ethereum-ad44cf28fb20).

The currently uploaded site doesn't include Trezor support as it won't work due to URL restrictions.

The reason I've included `http` next to `https` is:
- it is a fully client side wallet
- Ethereum RPC nodes do not support `https` at the moment. A tunnel/proxy needs to be used. Many users prefer to use their own nodes, but won't have `https` support.

Let me know if this is acceptable

### DIFF
--- a/signer/config.json
+++ b/signer/config.json
@@ -10,7 +10,9 @@
 		"https://coinprism\\.com(/.*)?",
 		"https://[\\w\\.-]+\\.coinprism\\.com(/.*)?",
 		"https://bitex\\.la(/.*)?",
-		"https://[\\w\\.-]+\\.bitex\\.la(/.*)?"
+		"https://[\\w\\.-]+\\.bitex\\.la(/.*)?",
+		"https://quorumwallet\\.com(/.*)?",
+		"https://[\\w\\.-]+\\.quorumwallet\\.com(/.*)?"
 		],
 	"blacklist_urls": [
 	],


runn1ng:
ping @prusnak

prusnak:
I really don't like adding http sites to the whitelist, but I understand the reason.

Any ETA for https support in Ethereum RPC? How hard is it for the users to configure their own node to use https tunnel/proxy?

I'll let @slush0 to have the final word.

axic:
@prusnak I do not think https support is considered/planned at all. I think they aim to move users to other interfaces designed with this use case in mind, however there's nothing there yet. (Apart from a dev preview from November called Mist.)

It is a standard JSON 2.0 RPC over HTTP. That means any HTTPS proxy can be used. Tech savvy users can do that with `stunnel`, but the majority of non-developers won't have a clue. All manuals state the RPC should listen on localhost only.

Transactions are prepared & signed in the browser, the RPC node only gets the public data which will be part of the blockchain.

I personally hope that the `http` part can be eliminated later this year, but not having it at the start can cripple adoption.

slush0:
You're talking about localhost RPC, but those are already whitelisted on HTTP. Your pull request is about http on quorimwallet.com, which have valid https already. So why not to use HTTPS on quorumwallet.com? Am I missing anything?

axic:
There are two RPC endpoints the site needs to connect to:
- Trezor
- Ethereum

The Ethereum RPC node (including the users local node) only runs on HTTP on the localhost. When running a website at HTTPS, connecting to HTTP (even if its localhost) is disallowed.

slush0:
Thank you, now I see the problem. Proper solution is to upgrade Ethereum RPC to HTTPS, not to downgrade quorumwallet.com to HTTP. We had to solve this kind of issues in mytrezor.com too; Trezor Bridge is using https, although it is on localhost, so HTTPS there does not add anyhing except solving cross-HTTP/HTTPS problems in browser. I recommend you to go this way, too, and tell users how to use stunnel and/or submit pull request adding HTTPS directly to Ethereum client.

slush0:
If I understand well, using quorumwallet over HTTP exposes MITM risk for users, which is especially sensitive in this kind of applications. So HTTPS is not just about nice green url bar...

axic:
There is a MITM risk to a certain level, although less with connecting to localhost. If they don't plan to use their own nodes, they can use the https site.

There are multiple node implementations (none of them supports https), but the most popular one plans to solve this with IPC and including a browser to run these apps within the "node environment". That however isn't available yet, not sure they really want to change anything on RPC given this roadmap.

Can we progress with enabling https for the time being?

slush0:
Sure, let's edit pull request to use https only, I'll accept it.

axic:
The change is done.

(Meanwhile I've got a response that they do not want to support HTTPS due to the complexity of it.)

slush0:
Who is "they"? Ethereum RPC?

axic:
The Go Ethereum (`geth`) implementation, which is the most widely used.

axic:
Thank you for merging. Can you please let me know when the signed configs are updated? Would it happen soon or it adheres to a fixed release schedule?

slush0:
Signed & added to https://github.com/trezor/webwallet-data. It will go out to official mytrezor.com and S3 locations with next release.

PR #12, opened 2016-06-10
Add support for ECDH session key generation (according to SLIP-0017)

romanz:
These definitions will be used to add ECDH support for GPG 2.1+ (following trezor/trezor-mcu#85).


prusnak:
Thanks!

PR #15, opened 2016-07-10
Prepare protocol for segwit.

jhoenicke:
These are the changes we need to support segwit.

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -291,6 +291,7 @@ message GetAddress {
 	optional string coin_name = 2 [default='Bitcoin'];
 	optional bool show_display = 3;			// optionally show on display before sending the result
 	optional MultisigRedeemScriptType multisig = 4;	// Filled if we are showing a multisig address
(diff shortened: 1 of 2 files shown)


prusnak:
Thanks!

ecdsa:
@jhoenicke can Trezor send to native segwit outputs? 
I get "RuntimeError: Failed to compile output" when trying with the latest firmware
(I set  txoutputtype.address to the bech32 serialized address, not sure if that is what trezor expects)

saleemrashid:
@ecdsa Not at the moment https://github.com/trezor/trezor-mcu/blob/v1.5.2/firmware/transaction.c#L166

prusnak:
Yes, native segwit is disabled in TREZOR for the time being.

jhoenicke:
I have a branch with bech32 support, but it is not yet finished.  I plan to support it with PAYTOADDRESS. The address should be autodetected.

ecdsa:
thank you for the clarification

PR #17, opened 2016-09-05
Matrix Recovery

jhoenicke:
This is the communication part of my matrix recovery patch.  It adds two new optional fields to the existing protocol messages for recovery:
- In message `RecoveryDevice` the enum `RecoveryDeviceType` is added.  This is a way for the PC to tell the Trezor which recovery type it prefers.  If this is not set it defaults to the old scrambled words mechanism that every PC must support for backward compatibility.
- In message `WordAck` there is a new field WordRequestType that is set by Trezor.  If the PC indicated support for matrix recovery, TREZOR asks for Matrix9 or Matrix6 requests instead of word requests and then expects instead of a word a string containing a single digit (corresponding to the matrix element).  Thus, the Trezor decides which recovery method to use and it does so using this field.  An old Trezor will not set the field indicating a plain word request.

The protocol is designed such that it is fully backward compatible.  An old PC client can recover a new Trezor and a new PC client can recover an old Trezor without any complicated protocol detection. **Question**: Should we make the field RecoveryDeviceType a bit-field, so that the PC can explicitly state which recovery methods it supports?  Or do we assume the PC will always support all methods up to the method that it indicates as the preferred?  

See issue trezor/trezor-mcu#96 for a description of matrix recovery.

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -387,6 +387,8 @@ message RecoveryDevice {
 	optional string language = 4 [default='english'];	// device language
 	optional string label = 5;				// device label
 	optional bool enforce_wordlist = 6;			// enforce BIP-39 wordlist during the process
+	// 7 reserved for keepkey recovery method
+	optional uint32 type = 8;				// supported recovery type (see RecoveryType)
 }
 
 /**
@@ -396,6 +398,7 @@ message RecoveryDevice {
  * @prev WordAck
  */
 message WordRequest {
+	optional WordRequestType type = 1;
 }
 
 /**
--- a/protob/types.proto
+++ b/protob/types.proto
@@ -105,6 +105,32 @@ enum PinMatrixRequestType {
 }
 
 /**
+ * Type of recovery procedure. These should be used as bitmask, e.g.,
+ * `RecoveryDeviceType_ScrambledWords | RecoveryDeviceType_Matrix`
+ * listing every method supported by the host computer.
+ *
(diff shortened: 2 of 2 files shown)


protob/messages.proto
```diff
@@ -387,6 +387,7 @@ message RecoveryDevice {
 	optional string language = 4 [default='english'];	// device language
 	optional string label = 5;				// device label
 	optional bool enforce_wordlist = 6;			// enforce BIP-39 wordlist during the process
+	optional RecoveryDeviceType type = 7;			// recovery type (scrambled words/matrix)
```
bgok: Field number 7 conflicts with a field number assigned by KeepKey. Would you consider using a different field number to avoid the conflict? As a maintainer of 2 third party wallet integrations (Multibit and a chrome client), I'm concerned that it will complicate integration if a common protobuf layout can't be used for both devices. 
See https://github.com/keepkey/device-protocol/blob/master/messages.proto#L362

jhoenicke:
I changed the field id to 8 as suggested by @bgok. I also made it a bitfield instead of an enum (still using the same enum for the constants but using an uint32 for the field as suggested here: http://stackoverflow.com/questions/26810188/protocol-buffers-and-enums-combinations

bgok:
Thanks!

prusnak:
Merged manually in https://github.com/trezor/trezor-common/commit/61af3d5e931ab41f0b81f462fbdc626d6bbec4f4

Thanks Jochen!

PR #19, opened 2016-10-16
Steem integration

xeroc:
This pull requests adds the basic Steem messages to trezor-common.

The trezor can be called with `SteemGetPublicKey` and `SteemSignTx` and either returns the PublicKey or a Signature message.
The `SteemSignTx` has an optional field for carring `transfers` for now only, but we will add more operation types over time. 
Since the used version of nanopb does not support `OneOf`, we decided to go with `optional` keywords and restrict ourselves to transactions that only carry **a single operation** even though the protocol/blockchain could deal with more.

These Operations are Steem specific and BitShares messages will be added separately when we get there. Some code will look similar.

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -854,14 +854,15 @@ message SteemOperationTransfer {
 }
 
 message SteemSignTx {
-	required uint32 ref_block_num = 1;
-	required uint32 ref_block_prefix = 2;
-	required uint32 expiration = 3;
+	repeated uint32 pubkey_n = 1;			// BIP-32 path to derive the key from master node
+	required uint32 ref_block_num = 2;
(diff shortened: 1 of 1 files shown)


xeroc:
I just realized that I forgot to add the `HDNode` to the `signTransfer` message. This way, the trezor can deal with multiple accounts on multiple networks. A specification will we published shortly in the SLIP repository.

xeroc:
The latest commit adds a `pubkey_n` to the calls similar to `address_n` to specify the actually used key ..
Since graphene doesn't deal with address **at all**, we use the terminology *pubkeys*

matejcik:
i'm pretty sure we can safely close this after more than two years :)

PR #23, opened 2017-01-22
Ethereum: chain id for EIP-155 replay protection

jhoenicke:
Added a field chain_id.  To get a backwards compatible signature
this field should not be set.  Otherwise it should be set to the
EIP-155 chain id.  Currently only chain id between 1 and 109 are
supported.
see trezor/trezor-mcu#142.


prusnak:
Thanks!

PR #36, opened 2017-07-24
NEM Support

saleemrashid:


saleemrashid:
Need to rebase and remove some fields, do not merge yet.

saleemrashid:
@prusnak Rebased and squashed

prusnak:
Any reason not to use call messages `NEMSignTx` and `NEMSignedTx` rather than `NEMTransaction` and `NEMRequestAnnounce` to follow the suit?

saleemrashid:
`NEMRequestAnnounce` refers to the request announce API method of NIS (e.g. the response from the TREZOR is normally sent verbatim to NIS to broadcast it).

`NEMTransactionCommon` is because when encoding the transaction, that part is referred to as "the common part of the transaction".

prusnak:
OK. And how about renaming `NEMTransaction` to `NEMSignTx`?

saleemrashid:
@prusnak If you want, you can rename it when you merge it and I'll change it everywhere else

prusnak:
It would be nice. I know we are not consistent everywhere, but at least we try to name input messages form <Verb><Noun> and output messages in form [Adjective]<Noun>.

saleemrashid:
OK. When you merge it, you can rename it to `NEMSignTx` and `NEMSignedTx`.

protob/types.proto
```diff
@@ -270,3 +270,137 @@ message IdentityType {
...
+
+/**
+ * Structure representing the mosaic supply change part for NEM transactions
+ * @used_in NEMTransaction
+ */
+message NEMMosaicSupplyChange {
+	optional string namespace = 1;			// Fully qualified name of the namespace
+	optional string mosaic = 2;			// Name of the mosaic definition
+	optional NEMSupplyChangeType type = 3;		// Type of supply change
+	optional uint64 delta = 4;			// Supply delta
```
prusnak: Don't you want to use sint64 for delta like you use sint32 below for NEMAggregateModification? (And we can drop the NEMSupplyChangeType). Just asking ...

saleemrashid: It's simpler this way, because I am keeping it the way it is serialized (`relative_change` is serialized as a signed integer, `delta` is serialized as unsigned with another field for the supply change type)

prusnak: Makes sense.

prusnak:
Thank you! Merged in. Also added https://github.com/trezor/trezor-common/commit/0e3f6e73ed1d97c653fea162581d47a74d0c2854 which renames the messages like said above.

saleemrashid:
Will update trezor/trezor-mcu#196 and open pull requests against python-trezor, TREZOR Connect and trezor.js

prusnak:
Thank you, much appreciated!

PR #39, opened 2017-08-06
Add 0xproject.com to domain whitelist

fabioberger:


prusnak:
Where can we test whether the integration works as expected?

prusnak:
Thanks!

PR #43, opened 2017-09-18
udev compat with all major linux distros

lrvick:
Virtually no linux distirbutions rely on the 'dialout' group anymore. Some older distros still use the "plugdev" group but most have been switching to a convention of using udev tags for device permissions instead.

This should avoid users having to create a group, add themseves to it, and logout/login. Plugdev group method still in place for backwards compat. Won't hurt anything.

I have been modifying this by hand for multiple systems of mine and others. Finally decided to PR it.

### DIFF
--- a/udev/51-trezor.rules
+++ b/udev/51-trezor.rules
@@ -2,18 +2,16 @@
 # https://trezor.io/
 # Put this file into /usr/lib/udev/rules.d
 
-# Raspbian does not understand the new TAG+="uaccess", TAG+="udev-acl" syntax, use MODE+GROUP for now ... :-(
-
 # TREZOR
-SUBSYSTEM=="usb", ATTR{idVendor}=="534c", ATTR{idProduct}=="0001", MODE="0666", GROUP="dialout", SYMLINK+="trezor%n"
-KERNEL=="hidraw*", ATTRS{idVendor}=="534c", ATTRS{idProduct}=="0001",  MODE="0666", GROUP="dialout"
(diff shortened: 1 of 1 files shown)


prusnak:
Could you confirm the proposed solution works on Raspbian as intended?

prusnak:
Also this [issue](https://bugzilla.redhat.com/show_bug.cgi?id=859244) says `plugdev` group is not present in Fedora. Could you confirm the proposed solution works on non-debian distros as well?

lrvick:
The "plugdev" group  is mainly there for compat with older debian based distributions. Raspbian is still one of the few that still uses it. Most distros now use the "uaccess" tag and a few use the older "udev-acl" tag.

By granting access via all 3 methods we ensure maximum compatibility as each distro uses -one- of these methods and will ignore the other two.

To be extra safe I just tested this on distro that lacks a plugdev group and relies on tags (arch) and one that relies on the plugdev group still (raspbian) and both work with the same rules file as expected.

Combining methods for compatibility is a common suggestion for other devices such as yubikey: https://wiki.debian.org/Smartcards/YubiKey4

prusnak:
Thanks!

saleemrashid:
After reading this issue, I wondered why the old rules work on my Arch Linux installation (since it doesn't use `dialout`) and I realised that the mode is set to `0666`. Perhaps this should be made stricter since it negates the benefits of `TAG+="uaccess"`?

prusnak:
Ah, right.

@lrvick could you try changing `0666` to `0660` and try whether that works for you on all distros you tried?

lrvick:
Great catch @saleemrashid 

@prusnak confirmed 0660 worked as expected in each target and made #45

lrvick:
Also, 0660 happens to be udev default if no mode is specified, but it does not hurt to be explicit.

PR #44, opened 2017-09-18
change BCH address prefix

michal-novacek:


prusnak:
ACK https://github.com/Bitcoin-ABC/bitcoin-abc/issues/68

PR #47, opened 2017-10-12
added bitcore urls to coin.json + renamed BCash

szymonlesisz:


prusnak:
We cannot rename "Bcash" to "Bitcoin Cash" because this is the identifier used to identify the coin in the firmware and this cannot be changed without a firmware update. I suggest adding "coin_label" or something similar to the structure if you want to use different string in the UI than the one used in "coin_name".

szymonlesisz:
Sure, i didn't know that, i thought it was just some leftover.
New field "coin_label" added

prusnak:
👍

PR #48, opened 2017-10-17
Adding Bitcoin 2X

runn1ng:
We will probably need b2x for mytrezor + connect in coins.json.

This is an idea, how to do it.

I realize, that having "B2X" is problematic, because there is no such thing as "only sending B2X" because of the lack of replay protection.

However, it's also problematic to have two currencies, both with "BTC".

And doing "BC1" and "BC2" would complicate everything even more. So I don't know how to get out of this.

### DIFF
--- a/coins.json
+++ b/coins.json
@@ -282,5 +282,34 @@
 	"min_address_length": 35,
 	"max_address_length": 95,
 	"bitcore": []
-}
-]
+}, {
+	"coin_name": "Bitcoin",
+	"coin_shortcut": "BTC",
+	"coin_label": "Bitcoin Segwit2x",
+	"address_type": 0,
+	"address_type_p2sh": 5,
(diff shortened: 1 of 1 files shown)


prusnak:
1) `firmware` field should be set to null, because we don't want to create another coin in the firmware.

2) also we would need to set `coin_name` to `Bitcoin` so the original coin is used in the firmware, or have a "hack" in the webwallet to send "Bitcoin" if "Bitcoin2x" was encountered

runn1ng:
I have edited the coins.json file.

It is still not working without hacks in trezor web wallet right now because of how localstorage works (uses coin shortcut), but I fixed that by a temporary hack. 

Trezor web wallet doesn't use the label currently.

cc @szymonlesisz I am not sure if this causes any problems in connect, but it shouldn't

PR #49, opened 2017-10-22
Add *.unchained-capital.com to domain whitelist

dhruvbansal:
Our website (https://www.unchained-capital.com) uses trezor.js to

* embed interactions with Trezor devices into our application's workflows
* support multiple coins, testnets and our own custom transaction types and BIP32 derivation paths
* prevent leaking any information about our transactions to third parties (including trezor.io) before broadcasting them to their blockchains

For these reasons, we were unable to use TrezorConnect.

Thanks for your great work.  We love our Trezors!


prusnak:
Will be released to production soon.

PR #57, opened 2017-11-23
Add null to config, to fix chrome apps + electron + nw.js

karel-3d:
Rehashing PR #20 

Chrome apps, Electron apps, nw.js apps set origin as "null". For that reason, our own password manager stopped working with bridge.

We originally merged #20 but reverted it for security issues. I think now it's time to merge it, since nw.js apps require this, and Copay will on desktop work as nw.js app only.


runn1ng:
I also removed the chrome-app URL, since it is useless now. 

I don't understand the failing Travis test

prusnak:
Do we still need this?

karel-3d:
Probably not, closing for now

PR #58, opened 2017-11-29
Vertcoin support

jhoenicke:
I looked at what we need for Vertcoin support and it's really not that much.  I'm not sure about minfee/maxfee/dust limit setting.  All other settings should be okay (I copied the genesis hex from the full node sources, hope this is correct).  Vertcoin uses the same 3... addresses as bitcoin for P2SH.

Of course, you need to increase the coin count in trezor-mcu when this is added.

It's possible to use it with electrum, see https://github.com/vertcoin/electrum-vtc/pull/47.

### DIFF
--- a/coins.json
+++ b/coins.json
@@ -439,5 +439,34 @@
 	"bitcore": [
 		"http://explorer.fujicoin.org/"
 	]
+},
+{
+	"coin_name": "Vertcoin",
+	"coin_shortcut": "VTC",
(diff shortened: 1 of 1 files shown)


prusnak:
LGTM :+1:

prusnak:
trezor-mcu updated via https://github.com/trezor/trezor-mcu/commit/f44635a9cafc7682f1f80dbdbb6645d1070ce552

PR #60, opened 2017-12-01
Removing digiexplorer

runn1ng:
Digiexplorer's API is not working

Websocket starts, but the query `getBlockHeader` times out.


runn1ng:
I am not sure if we should link to external explorers here that we have no control of and that will probably barely work.

saleemrashid:
> Websocket starts, but the query getBlockHeader times out.

From what I can tell, this is because old versions of Insight do not support the `message` channel.

PR #61, opened 2017-12-02
Messages for iota communication

bartslinger:
Required for upcoming pull-request in trezor-mcu

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -84,16 +84,13 @@ enum MessageType {
 	MessageType_CosiCommitment = 72 [(wire_out) = true];
 	MessageType_CosiSign = 73 [(wire_in) = true];
 	MessageType_CosiSignature = 74 [(wire_out) = true];
-	MessageType_IotaShowSeed = 75 [(wire_in) = true];
-	MessageType_IotaGetAddressCounter = 76 [(wire_in) = true];
-	MessageType_IotaAddressCounter = 77 [(wire_out) = true];
-	MessageType_IotaSetAddressCounter = 78 [(wire_in) = true];
-	MessageType_IotaGetAddress = 79 [(wire_in) = true];
-	MessageType_IotaAddress = 80 [(wire_out) = true];
-	MessageType_IotaTxRequest = 81 [(wire_in) = true];
-	MessageType_IotaTxApproved = 82 [(wire_out) = true];
-	MessageType_IotaTxDetails = 83 [(wire_in) = true];
-	MessageType_IotaSignedTx = 84 [(wire_out) = true];
+	MessageType_IotaGetAddressCounter = 75 [(wire_in) = true];
+	MessageType_IotaAddressCounter = 76 [(wire_out) = true];
+	MessageType_IotaSetAddressCounter = 77 [(wire_in) = true];
+	MessageType_IotaGetAddress = 78 [(wire_in) = true];
+	MessageType_IotaAddress = 79 [(wire_out) = true];
+	MessageType_IotaTxRequest = 80 [(wire_in) = true];
+	MessageType_IotaSignedTx = 81 [(wire_out) = true];
 	MessageType_DebugLinkDecision = 100 [(wire_debug_in) = true, (wire_tiny) = true];
 	MessageType_DebugLinkGetState = 101 [(wire_debug_in) = true];
 	MessageType_DebugLinkState = 102 [(wire_debug_out) = true];
@@ -928,7 +925,7 @@ message CosiSignature {
 
 /**
  * Request: Get address counter
- * @next IotaGetAddressCounter
+ * @next IotaAddressCounter
  */
 message IotaGetAddressCounter {
 }
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -927,14 +927,6 @@ message CosiSignature {
 ///////////////////
 
 /**
- * Request: Show seed on the device
- * @next Success
- * @next Failure
- */
-message IotaShowSeed {
-}
-
-/**
  * Request: Get address counter
  * @next IotaGetAddressCounter
  */
@@ -946,7 +938,7 @@ message IotaGetAddressCounter {
  * @prev IotaGetAddressCounter
  */
 message IotaAddressCounter {
-	required uint32 address_counter = 1;
+	optional uint32 address_counter = 1;
 }
 
 /**
@@ -955,7 +947,7 @@ message IotaAddressCounter {
  * @next Failure
  */
 message IotaSetAddressCounter {
-	required uint32 address_counter = 1;
+	optional uint32 address_counter = 1;
 }
 
 /**
@@ -971,8 +963,8 @@ message IotaGetAddress {
  * @prev IotaGetAddress
  */
 message IotaAddress {
-	required uint32 address_index = 1;
-	required string address = 2;
+	optional uint32 address_index = 1;
+	optional string address = 2;
 }
 
 /**
@@ -981,10 +973,10 @@ message IotaAddress {
  * @next Failure
  */
 message IotaTxRequest {
-	required string receiving_address = 1;			// Address to which the amount is send
-	required uint64 transfer_amount = 2;			// Amount of iota to be transferred
-	required uint64 balance = 3;					// Balance on the current address (to calculate remainder)
-    required uint64 request_timestamp = 4;			// Final transaction must be within certain time from here
+	optional string receiving_address = 1;			// Address to which the amount is send
+	optional uint64 transfer_amount = 2;			// Amount of iota to be transferred
+	optional uint64 balance = 3;					// Balance on the current address (to calculate remainder)
+	optional uint64 request_timestamp = 4;			// Final transaction must be within certain time from here
 	optional uint32 input_address_index = 5;		// Seed index from where to sent
 	optional uint32 remainder_address_index = 6;	// Seed index for the remaining balance
 	optional string tag = 7;						// Transaction tag, defaults to 'TREZOR'
@@ -1005,7 +997,7 @@ message IotaTxApproved {
  * @next Failure
  */
 message IotaTxDetails {
-	required uint64 transaction_timestamp = 1;		// Timestamp of the transactions in the bundle
+	optional uint64 transaction_timestamp = 1;		// Timestamp of the transactions in the bundle
 }
 
 /**
@@ -1013,8 +1005,8 @@ message IotaTxDetails {
  * @prev IotaTxDetails
  */
 message IotaSignedTx {
-	required string bundlehash = 1;				// Bundle hash (hash of all transaction elements)
-	required string first_signature = 2;		// First signature (security level 1)
+	optional string bundle_hash = 1;			// Bundle hash (hash of all transaction elements)
+	optional string first_signature = 2;		// First signature (security level 1)
 	optional string second_signature = 3;		// Second signature (security level 2)		
 }
 
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -969,40 +969,22 @@ message IotaAddress {
 
 /**
  * Request: Approve a value transfer.
- * @next IotaTxApproved
+ * @next IotaSignedTx
  * @next Failure
  */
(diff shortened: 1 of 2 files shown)


protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+///////////////////
+// Iota messages //
+///////////////////
+
+/**
+ * Request: Show seed on the device
+ * @next Success
+ * @next Failure
+ */
+message IotaShowSeed {
```
saleemrashid: TREZOR never shows the seed after it is backed up by the user.

protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+}
+
+/**
+ * Request: Get address counter
+ * @next IotaGetAddressCounter
+ */
+message IotaGetAddressCounter {
```
saleemrashid: It should be the role of the wallet to store this information, not the device. Unless there is something I am missing.

protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+/**
+ * Request: Approve a value transfer.
+ * @next IotaTxApproved
+ * @next Failure
+ */
+message IotaTxRequest {
+	required string receiving_address = 1;			// Address to which the amount is send
+	required uint64 transfer_amount = 2;			// Amount of iota to be transferred
+	required uint64 balance = 3;					// Balance on the current address (to calculate remainder)
+    required uint64 request_timestamp = 4;			// Final transaction must be within certain time from here
```
saleemrashid: Indentation

protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+ */
+message IotaTxApproved {
+}
+
+/**
+ * Request: Send timestamp to initiate transaction signing.
+ * @next IotSignedTx
+ * @next Failure
+ */
+message IotaTxDetails {
```
saleemrashid: Why does this need to be separate from `IotaTxRequest`?

protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+	required uint64 transaction_timestamp = 1;		// Timestamp of the transactions in the bundle
+}
+
+/**
+ * Response: Transaction hash and signatures
+ * @prev IotaTxDetails
+ */
+message IotaSignedTx {
+	required string bundlehash = 1;				// Bundle hash (hash of all transaction elements)
```
saleemrashid: Why not `bundle_hash`?

protob/messages.proto
```diff
@@ -912,6 +922,102 @@ message CosiSignature {
...
+ */
+message IotaGetAddressCounter {
+}
+
+/**
+ * Returns the address counter as currently set in storage
+ * @prev IotaGetAddressCounter
+ */
+message IotaAddressCounter {
+	required uint32 address_counter = 1;
```
saleemrashid: Never use `required`, always use `optional`.

protob/messages.proto
bartslinger: This is not a trezor seed, but the IOTA seed. As long as I don't have a computer program that converts 24 words into the IOTA seed the same way, I'd like to display the seed on the screen.

protob/messages.proto
bartslinger: The problem is, when you switch to another computer, it might take a very very long time to log in. Every public address takes 12 seconds to generate. For every transaction ever done, the trezor needs to generate an address and check if there is balance on that address. For a small sacrifice in memory, this process can be much smoother.

protob/messages.proto
bartslinger: Is there a specific reason for that?

protob/messages.proto
bartslinger: Tried to explain it in https://github.com/trezor/trezor-mcu/pull/255

protob/messages.proto
bartslinger: No specific reason

protob/messages.proto
saleemrashid: Then I think it is a good idea to write a program to convert the mnemonic into the IOTA seed because the TREZOR never gives out private keys after initial backup

protob/messages.proto
bartslinger: Will fix it

protob/messages.proto
saleemrashid: How can the user restore a new device then? It would take far too long.

protob/messages.proto
saleemrashid: Explained here https://developers.google.com/protocol-buffers/docs/proto#specifying-field-rules

Furthermore, in Proto3, all fields are `optional`

protob/messages.proto
bartslinger: The address counter is known to the computer and can be saved. It can be restored with IotaSetAddressCounter. If the device is completely lost and the address counter is unknown, it takes up to a few minutes to restore the device, depending on how many transactions have been done with it. It basically then has to generate every single address from index 0 and check against the network if it has a balance.

protob/messages.proto
bartslinger: In that case my messages are not the only ones that don't comply. Anyway, I'll change them to optional.

protob/messages.proto
bartslinger: It's on the todo list. If it blocks merging, I'll remove the option.

prusnak: I guess it does not make sense to merge in the message which will be removed afterward. So let's remove the message.

protob/messages.proto
prusnak: Does the counter have to be sequential (1, 2, 3) or is monotonic (1, 12, 250, 100382) OK?

saleemrashid: Is there no way to improve the performance of address generation?

protob/messages.proto
prusnak: Please use `bundle_hash`

protob/messages.proto
saleemrashid: The old messages that do not comply unfortunately cannot be changed (for same reason stated in the link! :smile:)

protob/messages.proto
bartslinger: If the counter is sequential, it is in line with current wallet implementations. Basically it can be anything as long as no numbers are re-used. Sequential also makes recovery faster, compared to monotonic.

 Address generation cannot be done faster. It has to do with the completely different cryptography used, compared to other coins.

protob/messages.proto
bartslinger: will do

protob/messages.proto
bartslinger: Ok

protob/messages.proto
prusnak: > Every public address takes 12 seconds to generate. 

12 seconds on which spec? Desktop PC?

bartslinger: It's on the TREZOR. Desktop can derive it, if you expose the private key :-p But that defeats the whole use of a TREZOR.

prusnak: That's really unfortunate for HW wallet usecase that you cannot derive the next address without the knowledge of the private key :-/

prusnak:
What is the required accuracy of signature timestamp? Do we really need to do the extra round-trip to ask for a new timestamp? What happens if the timestamp is in the future?

bartslinger:
Good question. At this moment, an accurate timestamp is not enforced at all. I do see benefit from not having the round-trip. Perhaps it is better to just skip that for now and apply an update if/when it really becomes important.

prusnak:
Please do remove that. We can always make things more complicated, but let's have them simple in the beginning.

bartslinger:
Alright sounds like a plan. I'll just pick a timestamp that's a little bit into the future. Can always delay the broadcast moment.

prusnak:
Great!

Also, I just noticed that you return duplicate `IotaAddress.address_index` field. I don't think it should be filled in the response because it is a part of the `IotaGetAddress` request.

bartslinger:
The idea was to have it optional. If you send IotaGetAddress without the requested index, it uses the index (address_counter) that's in storage.

The alternative is to change the request sequence. Then, the computer would first need to send out IotaGetAddressCounter. Let me know what you prefer.

saleemrashid:
@bartslinger Correct me if I'm wrong but doesn't the type of signature IOTA uses make address reuse fatal? (e.g. exposes the private key). If so, that's a major attack vector (because malware can change address index and expose the private key)

prusnak:
@saleemrashid Do I understand it correctly that you suspect that using `IotaTxRequest` two times in a row with the same fields just with a changed `tag` field (for example) would result in having two different `IotaSignedTx` results which can be used to compute the used private key?

bartslinger:
Forging multiple signatures using different data and from the same address is bad. Just generating an address is not a problem if it is done multiple times.

To prevent this:
* The trezor increments the address counter when a signed transaction is sent out
* The trezor requires a confirmation on the device to set the address counter to another value.

prusnak:
> The trezor increments the address counter when a signed transaction is sent out

But this counter is not enforced, user can override it by setting `IotaTxRequest.input_address_index`, no?

saleemrashid:
@prusnak That was my concern. Malware can trick the user into reusing an address and then use `input_address_index`.

bartslinger:
I forgot that I actually thought about that before. It requires an additional confirmation on the trezor if that field is used.

https://github.com/iota-trezor/trezor-mcu/blob/master/firmware/fsm.c#L1555

The only reason why I included this option is the following scenario: The network protocol (still) allows to send IOTA to an address that has already been sent from. I'm not sure if this possibility will ever be removed. If for some reason, a person receives IOTA on an old address, he might want to get it out of there. As far as I understand it, a second signature on the same address reduces security, but is not yet the end of the world.

If the user would not be able to select a custom index using `IotaTxRequest.input_address_index` and `IotaTxRequest.remainder_address_index`, the best he can do is use `IotaSetAddressCounter` and send all of it to another wallet. However with these optional fields, the user can select the old address as input address and the remainder address can be the address that currently holds the value.

However while writing this, I realise that it is also possible to use `IotaSetAddressCounter` to change to the old address, and then send all of the IOTA in there to the newest address. After that, the address counter can be restored. **So actually we can get rid of these fields :)**

saleemrashid:
@bartslinger The user is still expected to remember the address counter when restoring a new device or trust the computer.

bartslinger:
@saleemrashid Is that a question?

Technically the user doesn't have to remember when restoring, but then the process will be like this:

1. Counter starts at value 0
2. Computer requests public address at current address (12 seconds generation time on Trezor)
3. Computer checks value on the IOTA network corresponding to that address.
4. If there is no value, increment counter and continue with step 2.

saleemrashid:
> 3. Computer checks value on the IOTA network corresponding to that address.

Which requires you to trust the computer, doesn't it? (It could lie in order to force address reuse)

bartslinger:
That is very true. In that case, yes, the user will need to remember the address counter. I should probably also add something to display it on the trezor.

bartslinger:
@prusnak @saleemrashid 

I am happy with the messages in the current form. If you have no more suggestions, I can start modifying the trezor-mcu code to reflect all of the feedback given here. Please let me know if you are also satisfied with the messages.

Merging is not yet required. I actually think it is better to wait and do it together with trezor-mcu when it is ready.

prusnak:
@bartslinger Thank you! Yes, you can start adapting the trezor-mcu code.

saleemrashid:
ACK everything @prusnak said

protob/messages.proto
bartslinger: I'm not sure what makes you think otherwise, but this is exactly what I've done already. The entire post-snapshot address re-generation is not required at all.

stockermans: Ok, sorry, I didn't realise that was what you have done already, I must have misunderstood the previous comments. Sorry for my long post.

bartslinger:
https://github.com/trezor/trezor-mcu/pull/255#issuecomment-379535626

PR #72, opened 2018-01-04
Added cashaddr_prefix

jhoenicke:
For bitcoincash new unique address format.


saleemrashid:
Couldn't we reuse `bech32_prefix`?

saleemrashid:
Actually, my bad since you can use this to determine whether the coin supports Cashaddr.

prusnak:
I knew that once `:` is used as HRP separator (so it's confusing when used in URI scheme), they will stop using the prefix completely. See here for example: https://cashaddr.bitcoincash.org

This way the addresses of different network (cash, cash testnet, etc). become indistinguishable again. :-/

prusnak:
Btw, does Electron Cash support TREZOR?

jhoenicke:
> This way addresses of different networks (cash, cash testnet, etc). become indistinguishable again

The prefix should always be used if it is not clear from the context, e.g., in a donation address link you use it.  I'd suggest to also use it when displaying the receive address on the PC for copy&paste.  When sending, we should allow the user to omit it, but it must also work when it is included.

The long checksum makes sure that an address is unique, even without the prefix.  If adding the prefix gives a valid checksum, then the address is correct.  We should use the prefix for all internal messages though, e.g. for querying the blockexplorer or in the USB protocol.  Omitting the prefix should only be done when interacting with the user and it is clear from the context.

> Btw, does Electron Cash support TREZOR?

Yes, although we need to backport the recent patches to support new python-trezor, e.g., protobuf interface.  I have a working version of electron-cash.  If trezor-0.9.0 is stable now I can submit a pull request.

prusnak:
> If trezor-0.9.0 is stable now I can submit a pull request.

I guess we won't be touching communication API for now.

PR #79, opened 2018-02-19
Added DigiPulse project to whitelist.

diidiiman:
Digipulse will improve users security and UX simplicity by utilising toolset available from integrations with Trezor.

### DIFF
--- a/signer/config.json
+++ b/signer/config.json
@@ -17,6 +17,7 @@
 		"https://[\\w\\.-]+\\.dash\\.run(/.*)?",
 		"https://0xproject\\.com(/.*)?",
 		"https://[\\w\\.-]+\\.unchained-capital\\.com(/.*)?",
+		"https://app\\.digipulse\\.io(/.*)?",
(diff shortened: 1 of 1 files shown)


prusnak:
You should use TREZOR Connect for integrations with TREZOR: https://github.com/trezor/connect 

Whitelist will be obsoleted soon and using it is discouraged for new projects.

diidiiman:
Ah snap. We were really looking forward on creating a simplified integration (just like pass manager extension).
Ok, thanks for the heads up!

prusnak:
TBH integration using Connect is even easier, as it moves low-level application logic (such as PIN entering) to a separate environment and you don't have to deal with that.

diidiiman:
It is easier for the devs, but more difficult for the end user, since the window block, different visual env. etc. This is why we wanted to go this path instead of the simplified solution so the UX would not suffer.
But I guess with connect we are limited to the popup window, right?

prusnak:
There is a new version of Connect (v5) in progress, which can be used as React component, etc. so you won't be limited to popup window only.

diidiiman:
Ok, thanks for the heads up!

PR #80, opened 2018-02-19
coins: add litecoin testnet

untrustbank:
Hi, I'd like to add support for the Litecoin Testnet. I hope I did this correctly.
`xpub_magic_segwit_p2sh` is not defined in slip-0132 yet, so I left it null.
Also, the Litecoin project is in the middle of a relay fee reduction (it's in the master branch of their repo), but it's not widely deployed yet. That is why the testnet `minfee_kb` is lower than mainnet LTC.

### DIFF
--- a/coins.json
+++ b/coins.json
@@ -202,6 +202,39 @@
 	]
 },
 {
+	"coin_name": "Litecoin Testnet",
+	"coin_shortcut": "TLTC",
+	"coin_label": "Litecoin Testnet",
+	"curve_name": "secp256k1",
+	"address_type": 111,
+	"address_type_p2sh": 58,
+	"maxfee_kb": 40000000,
+	"minfee_kb": 1000,
+	"signed_message_header": "Litecoin Signed Message:\n",
+	"hash_genesis_block": "4966625a4b2851d9fdee139e56211a0d88575f59ed816ff5e6a63deb4e3e29a0",
+	"xprv_magic": "04358394",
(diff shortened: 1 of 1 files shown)


untrustbank:
To make this a little more fun, I've left some TLTC laying around if you need some for a quick test :)

```
private key cPnfjmGVBGyiDZnsHWMQ1zsv154WYErtwt9EQFj6NrNq7b4oCqmU
txout 11742cdf26199af4e55bc3b472e3b38fa7d655404757770e7511da529f770707 1
{
  "value": 460.59891258,
  "scriptPubKey": {
    "asm": "OP_DUP OP_HASH160 11b85711552112fd01fd7176f170fe2205367a9f OP_EQUALVERIFY OP_CHECKSIG",
    "hex": "76a91411b85711552112fd01fd7176f170fe2205367a9f88ac",
    "type": "pubkeyhash",
    "addresses": [
      "mh8ecVPxa6BfpnrKBqEN1MEXdqqpZGRuGj"
    ]
  },
}
```

untrustbank:
Noting that `https://testnet.litecore.io/` appears to be down. I guess that bitcore node will not be reliable. Maybe Trezor could spin up a Litecoin testnet node?

prusnak:
We won't be adding this to the official firmware. If Litecoin community has a stable backend, we might reconsider this.

untrustbank:
OK, thanks for the info.

PR #81, opened 2018-02-20
Adding support for Flashcoin

flashbountyhunter:


flashbountyhunter:
Is there any sort of timeline for when this pull request will be committed?  It would be nice to know when we can expect FLASH support to be enabled.

prusnak:
I am reworking the definitions and splitting the file `coins.json` into multiple smaller json files. These will reside in `defs/coins`, see https://github.com/trezor/trezor-common/commit/4fff5fbb7f3b184677e93e6cce320774835dc888

Your coin from this pull request was added in the commit above. Thanks!

srsflash:
Hi, we are wondering about the status and ETA of FLASH support as we don't see it listed on your website as 'soon' though this pull request was added in the commit above.

Thanks!!

PR #84, opened 2018-03-07
Changed tx_type to uint32

tyrion70:
As discussed, changed tx_type from bytes to uint32


prusnak:
Changed comment in 7e61f3703883da2f58d1988b373503cfe7c3561c

PR #86, opened 2018-03-12
Add Cryptomover domain to config.json

cm-abhijit:


prusnak:
You should use TREZOR Connect to interact with TREZOR from your website: https://github.com/trezor/connect

Config will be phased out soon.

cm-abhijit:
@prusnak We are using Trezor Connect Only. Please accept our domain name white list request.

prusnak:
No need to do a whitelist if you use Connect. It works without it.

PR #87, opened 2018-03-13
proto: add `auto_lock_delay_ms` field to ApplySettings and Features

romanz:
Following https://github.com/trezor/trezor-core/issues/163 (I am planning to add [this feature](https://github.com/romanz/trezor-mcu/tree/auto-lock) to TREZOR One).

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -175,6 +175,7 @@ message ApplySettings {
 	optional bool use_passphrase = 3;
 	optional bytes homescreen = 4;
 	optional PassphraseSourceType passphrase_source = 5;
+	optional uint32 auto_lock_delay_ms = 6;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


prusnak:
I am not sure whether we need to report the value in features.  Is there
any usecase when host needs to know the value?

romanz:
I have added it there mainly due to similarity to other settings (e.g. label and language), but it's not required for this feature.
Should I just remove it?

prusnak:
Yes. I would remove it. While label and language are required to be known by the host, this flag isn't.

Please add it to storage.proto, though. We don't use this file internally, but it's nice to keep track of stuff being stored in the storage.

romanz:
Sounds good :)
Updated `trezor-common`, `trezor-mcu` and `python-trezor` PRs.

saleemrashid:
Regarding returning the value in `Features`, isn't that useful for displaying the current value to the user? (i.e. in the Device Settings page)

romanz:
@prusnak I'm fine with both suggestions, please let me know what do you prefer :)

prusnak:
Let's not return that in Features for now. Thanks, Roman!

PR #95, opened 2018-04-12
Fix Lisk potobuf messages

alepop:
Change `amount` and `fee` from `string` to `uint64`
Change `sender_public_key`, `requester_public_key` and `singature` from `string` to `bytes`


prusnak:
Thanks!

PR #103, opened 2018-05-04
Add Koto

wo01:


prusnak:
`"max_address_length": 95` - please show me an address in insight which is 95 characters long

PR #105, opened 2018-05-11
Add Feathercoin

lclc:
Thanks! :)

Our Bitcore server should be up and running in ~1 week, but we plan to focus on using Trezor with our Electrum fork (https://github.com/Feathercoin-Foundation/electrum-ftc) as well as our upcoming new desktop wallet (supporting BTC, FTC, LTC and Trezor).


lclc:
Hey guys, any chance Feathercoin gets activated in upstream Trezor soon?

We have smaller exchanges and other service providers asking for Feathercoin support in Trezor

prusnak:
Did these smaller exchanges and other service providers integrated Trezor using Connect into their products?

rogerdarin:
> Did these smaller exchanges and other service providers integrated Trezor using Connect into their products?

The exchange operator wouldn't tell us what they're using themselves but they're clear that they want their customers to be safe when withdrawing their coins and they spell it out quite clearly (even on their website) that people keeping their coins 'at home' should put them on Trezor.

PR #110, opened 2018-05-22
Add Totle url to signer config

passabilities:


prusnak:
Please don't use trezor.js directly. We won't be adding new URLs to config. Use https://github.com/trezor/connect instead.

passabilities:
Why are no more URLs be added? We are trying to integrate our own UI instead of having a popup.

passabilities:
What is the reasoning behind this? The only reason I started working with the trezor.js library is so that we can have our own custom interface rather that having a popup window that interrupts the flow of our app.

passabilities:
:thinking: why does this warrant for shutting off access to the trezor.js library though?

prusnak:
You can use the trezor.js library. But not from the web (for security reasons). If you want to access the TREZOR from the web, you need to use Connect.

passabilities:
Sooo what your saying is, I can only use the trezor.js library if i build a desktop application that is not on the web? :weary:

prusnak:
yes

PR #112, opened 2018-05-25
ZenCash support

josephnicholas:
Good day all,

Please review for ZenCash support.
@prusnak @matejcik


josephnicholas:
Redoing implementation

PR #118, opened 2018-06-05
Add Zcoin to support.json

yura-pakhuchiy:


matejcik:
our support data is now up-to-date from a different source somehow (or at least will be when #173 is merged) so closing this

PR #122, opened 2018-06-08
Zencash support

josephnicholas:
Zencash integration for Trezor model T


prusnak:
Don't edit coins_details.json before your coin is supported in TREZOR.

Don't add new fields to defs/coins/coin.json without discussion. Field `maintainer` should contain valid email of the maintainer.

Don't add new fields to protob/types.proto without discussion.

PR #134, opened 2018-06-13
Zencash Integration

josephnicholas:
**Short Description from the original PR:**
Zencash has bip115 implementation as replay protection.
According bip115 as additional params we have to use block hash and block height. So we add 2 fields in message **TxOutputType** ( _block_hash_bip115 and block_height_bip115_ ).
Also we can't regenerate script_sig because we have to know block_hash and block_height of transaction which we wanna spend, so was added one more field in message **TxInputType** (_prev_input_script_).


prusnak:
What is the length of `prev_input_script`? Isn't it better to just have `block_hash_bip115` and `block_height_bip115` in `TxInputType` too?

PR #137, opened 2018-06-13
Adding balance.io to whitelist for Trezor integration

jinchung:


prusnak:
We don't add new entries to whitelist. Use https://github.com/trezor/connect for Trezor integration into your website.

PR #138, opened 2018-06-15
Using new FW for BGold

karel-3d:
In current FW and current wallet, we wrongly use "Bitcoin" coin name for BGold.

Only in new firmware we accept "BGold" as coin name, so I bumped the version here


prusnak:
Fixed for both T1 and T2 in https://github.com/trezor/trezor-common/commit/47640c6828603951e619ba9d327c2c080d4072aa

PR #140, opened 2018-06-18
Zencash support

josephnicholas:
Hi @prusnak based on the previous PR, you have asked the following questions.

> What is the length of prev_input_script? Isn't it better to just have block_hash_bip115 and block_height_bip115 in TxInputType too and reconstruct it using that data?

On this implementation, the length of the previous input script when replay protection is used is 63. 

Adding block_hash and block_height as fields of the TxInput type will require parsing the input script, verifying if the input script contains the replay protection opcode, in such case extract the block_hash and block_height, pass them and finally recreate the script and then sign it.
Using the prev_input_script field approach would prevent such process.

@idimon4uk do you have any additions?


josephnicholas:
Hi @prusnak based on the previous PR, you have asked the following questions.

> What is the length of prev_input_script? Isn't it better to just have block_hash_bip115 and block_height_bip115 in TxInputType too and reconstruct it using that data?

On this implementation, the length of the previous input script is 63. I don't really suggest doing the block_hash and block_height as parameters of the TxInput type since we have to again reconstruct the script and parse it.

Another thing also, prev_input_script was being used since it has the all the script data the current transaction has, rather than reconstructing the data and parsing it again, I think just getting the prev_input_script is effecient.

@idimon4uk do you have any additions?

idimon4uk:
Also Zencash didn't  use replay protection (bip155 implementation) till fork from Zcash. Solution with  previous _input_script as field allow solve problem with spending very old transaction (transactions till fork).

prusnak:
What exactly is the contents of `prev_input_script`? Can you break this down into individual (63) bytes?

idimon4uk:
Pavol, thank you for your comment. Please let us try to briefly explain the logic behind our solution.

It is required to parse the input script to identify block_hash and block_height of the replay protection for the following reasons:
* not all the transactions contain replay protection (old transactions created before replay protection introduction)
* transactions with replay protection can specify different block_hash and block_height according to bip-0115 

If you prefer this "parse, extract, and reconstruct" approach instead of passing the script "as it is", we can implement in this way. 

Please let us know your thoughts. We are ready to follow your final recommendation.

prusnak:
I prefer "parse, extract, reconstruct" approach because phases 1 and 2 happen outside of the device. Also reconstructing the data is much simpler than parsing it and we prefer to keep firmware as simple as possible.

idimon4uk:
Hello Pavol,
We’ve made the updated pull request.
Could you please review it?
I appreciate your feedback.

prusnak:
Looks good now!

PR #142, opened 2018-06-21
Add Koto

wo01:


prusnak:
Koto uses same address_type for P2PKH and P2SH as Zcash. Please change these values first, otherwise this is really confusing and cannot be accepted.

I accepted the official site change in 2a3a7dec147e4688921991b1d5da820e223817c7

prusnak:
I see on your Insight that the prefixes are indeed different, so you should probably fix these values.

PR #144, opened 2018-06-23
Update version group id of Koto

wo01:
We have reverted SigHash personalization and updated version group id.
Koto Overwinter does not yet activated. We will release new Overwinter-compliant Koto in near future.
Could you re-enable Koto?

### DIFF
--- a/defs/coins/koto.json
+++ b/defs/coins/koto.json
@@ -23,7 +23,7 @@
   "decred": false,
   "fork_id": null,
   "force_bip143": false,
-  "version_group_id": null,
+  "version_group_id": 48748912,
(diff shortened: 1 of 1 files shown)


prusnak:
When will the Overwinter activate?

wo01:
The actual activation height is not yet decided.
We will decide the height immediately and then release new version of Koto in one or two weeks.
We plan to activate a month later after the release.

prusnak:
OK, the current firmware is frozen already, but we'll add Koto in the next iteration.

wo01:
All right.
Thanks.

PR #148, opened 2018-06-26
Add Cardano currency support

ddeath:
This PR is for adding Cardano currency support and is intended to merge with https://github.com/trezor/trezor-core/pull/250

We are NOT official developers of Cardano, however we are developers of [www.cardanolite.com](https://www.cardanolite.com)

### DIFF
--- a/protob/messages.proto
+++ b/protob/messages.proto
@@ -100,6 +100,21 @@ enum MessageType {
 	MessageType_DebugLinkMemoryWrite = 112 [(wire_debug_in) = true];
 	MessageType_DebugLinkFlashErase = 113 [(wire_debug_in) = true];
 
+
+	// Cardano
+	MessageType_CardanoSignMessage = 300 [(wire_in) = true];
+	MessageType_CardanoMessageSignature = 301 [(wire_out) = true];
+	MessageType_CardanoVerifyMessage = 302 [(wire_in) = true];
+	MessageType_CardanoSignTransaction = 303 [(wire_in) = true];
+	MessageType_CardanoTxRequest = 304 [(wire_out) = true];
+	MessageType_CardanoGetPublicKey = 305 [(wire_in) = true];
+	MessageType_CardanoPublicKey = 306 [(wire_out) = true];
+	MessageType_CardanoGetAddress = 307 [(wire_in) = true];
+	MessageType_CardanoAddress = 308 [(wire_out) = true];
+	MessageType_CardanoTxAck = 309 [(wire_in) = true];
+	MessageType_CardanoSignedTransaction = 310 [(wire_out) = true];
+	
+
 	// Lisk
 	MessageType_LiskGetAddress = 114 [(wire_in) = true];
 	MessageType_LiskAddress = 115 [(wire_out) = true];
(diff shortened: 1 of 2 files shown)


prusnak:
Thanks for your work. We'll review the pull request shortly. Once it is accepted, we can start reviewing other submitted stuff.

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+message CardanoAddresses {
+	repeated CardanoAddress addresses = 1;
+}
+
+/**
+ * Request: Ask device for Cardano addresses
+ * @next CardanoAddresses
+ * @next Failure
+ */
+ message CardanoGetAddresses {
```
prusnak: Let's drop this message and use `CardanoGetAddress` only.

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
 message DebugLinkFlashErase {
 	optional uint32 sector = 1;
 }
+
+/**
+ * Response: Contains a Cardano addresses derived from device private seed
+ * @prev CardanoGetAddresses
+ */
+message CardanoAddresses {
```
prusnak: Let's drop this message and use `CardanoGetAddress` only.

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+	optional bool show_display = 2;				// optionally prompt for confirmation on trezor display
+}
+
+/**
+ * Request: Ask device for Cardano address
+ * @next CardanoAddress
+ * @next Failure
+ */
+ message CardanoGetAddress {
+	required CardanoDerivationPath path = 1;		// BIP-32-style path to derive the key from master node
```
prusnak: Don't use required (it is obsoleted by protobuf3).

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+	repeated uint32 address_n = 1;				// BIP-32 path to derive the key from master node
+	required bytes message = 2;				// message to be signed
+}
+
+/**
+ * Request: Ask device for public key corresponding to address_n path
+ * @next CardanoXpubKey
+ * @next Failure
+ */
+ message CardanoGetXpubKey {
```
prusnak: Let's rename this to `CardanoGetPublicKey` to match the rest of the API

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+	repeated uint32 address_n = 1;				// BIP-32 path to derive the key from master node
+}
+
+/**
+ * Response: Contains public key derived from device private seed
+ * @prev CardanoGetXpubKey
+ */
+ message CardanoXpubKey {
```
prusnak: Let's rename this to `CardanoPublicKey`to match the rest of the API.

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+	required string xpub_key = 1;				// Xpub key
```
prusnak: Let's use ```

protob/messages.proto
```diff
@@ -1437,3 +1454,125 @@ message DebugLinkMemoryWrite {
...
+/**
+ * Request: Ask device to sign Cardano transaction
+ * @next CardanoSignedTransaction
+ * @next CardanoTxRequest
+ * @next Failure
+ */
+message CardanoSignTransaction {
+	repeated CardanoTxInputType inputs = 1;  		// inputs to be used in transaction
+	repeated CardanoTxOutputType outputs = 2;		// outputs to be used in transaction
+	repeated bytes attributes = 3; 				// transaction attributes
```
prusnak: What is the purpose of `attributes`? Can we use protobuf enum/submessage instead?

protob/types.proto
```diff
@@ -506,3 +506,42 @@ message LiskMultisignatureType {
...
+	required uint64 amount = 3;							// amount to spend
+}
+
+/**
+ * Structure representing cardano derivation path
+ * @used_in CardanoAddresses
+ */
+message CardanoAddress {
+	required string address = 1;						// Base58 address
+	repeated uint32 address_n = 2;						// BIP-32 path to derive the key from master node
```
prusnak: Let's move this into messages and drop `address_n` field.

protob/types.proto
```diff
@@ -506,3 +506,42 @@ message LiskMultisignatureType {
...
+}
+
+/**
+ * Structure representing cardano derivation path
+ * @used_in CardanoGetAddresses
+ * @used_in CardanoAddress
+ */
+message CardanoDerivationPath {
```
prusnak: Let's drop this.

protob/messages.proto
ddeath: Currently it is not defined and attributes must be empty. But cardano team is mentioning that in version 1 (current is 0), attributes will be used for additional data storage. More about it can be found [here](https://github.com/input-output-hk/cardano-sl/blob/8d25c2ad3ca2354af8f8c43a2972d1b9a31bf440/docs/tx/processing.md)

prusnak: Let's drop this attribute then. We can always add it later (preferably in form of a submessage, not arbitrary bytes value).

protob/messages.proto
ddeath: `root_hd_passphrase` is needed for address derivation. For example if you want to get 20 cardano addresses you can ask 20x trezor for an address or you can ask for pub key and root_hd_passphrase and derive it your self. Also root_hd_passphrase is needed for address decryption. Address decryption is a way how to tell if the address is our or not. You can read about it [here](https://github.com/input-output-hk/cardano-sl/blob/8d25c2ad3ca2354af8f8c43a2972d1b9a31bf440/docs/hd.md)

prusnak: OK, it seems that it is safe to share `root_hd_passphrase` in public. Let's keep this field + the two fields above (`HDNodeType node` and `string xpub`).

ddeath:
All comments should be fixed now.

protob/messages.proto
```diff
@@ -1437,3 +1452,113 @@ message DebugLinkMemoryWrite {
...
+	optional string root_hd_passphrase = 3;			// Hd passphrase for root in hex format
+}
+
+/**
+ * Request: Ask device to verify Cardano message
+ * @next Success
+ * @next Failure
+ */
+message CardanoVerifyMessage {
+	optional bytes public_key = 1;				// Public key which was used to sign message
```
prusnak: For other coins we use `address` not `public_key` in `VerifyMessage`, would it make sense to use it here too? (You don't return `public_key` but `address` in `CardanoMessageSignature` so it makes sense)

protob/types.proto
```diff
@@ -506,3 +506,24 @@ message LiskMultisignatureType {
 	optional uint32 life_time = 2;
 	repeated string keys_group = 3;
 }
+
+/**
+ * Structure representing cardano transaction input
+ * @used_in CardanoSignTransacion
+ */
+message CardanoTxInputType {
+	optional bytes tx_hash = 1;				// txHash
```
prusnak: Is it OK if we change the order and names of the fields to match `TxInputType`?

That is 

```
	repeated uint32 address_n = 1;						// BIP-32 path to derive the key from master node
	required bytes prev_hash = 2;						// hash of previous transaction output to spend by this input
	required uint32 prev_index = 3;						// index of previous output to spend
```

protob/messages.proto
prusnak: Ah, correct. Let's use `optional bytes public_key = 1;` in `CardanoMessageSignature` for consistency.

After this change, the messages `CardanoSignMessage`, `CardanoVerifyMessage` and `CardanoMessageSignature` will be exactly the same as `LiskSignMessage`, `LiskVerifyMessage` and `LiskMessageSignature`, which is another Ed25519 currency, so it makes sense

ddeath: fixed

prusnak:
Thank you!

ddeath:
I just discovered that there wont be an option to use trezord-go for comunication with trezor directly from a website and it have some implications. For example if the online wallet would like to get 10 addresses from trezor, in current situation it would mean 10 popups for user.

Could we bring back CardanoGetAddresses so we can obtain more addresses with only one popup?

prusnak:
You should be able to derive addresses using `CardanoGetPublicKey`, right?

prusnak:
I see. We are modifying Connect right now, so it will be possible to call 10x GetAddress without showing 10 popups for user (just the first one).

ddeath:
Ok, thank you.

tsusanka:
Just FYI I've renamed the CardanoSignedTransaction message to CardanoSignedTx to match the other coins https://github.com/trezor/trezor-common/commit/debd6871d672fc2b4b04c4e1390ac6ad209c6093

prusnak:
@tsusanka In that case, don't you want to rename `CardanoSignTransaction` to `CardanoSignTx` too?

tsusanka:
Sounds about right. Will do

ddeath:
@tsusanka no it is not used. The response is send with as `CardanoTxRequest`

tsusanka:
Ok, I'll remove it then completely

tsusanka:
Or do we want to modify this to return the result in CardanoSignedTx?

tsusanka:
I've modified the code in trezor-core to return CardanoSignedTx in e657397c83fed4bd5b02176f63547b120130dd56

PR #152, opened 2018-06-28
[WIP] Tron Support

Rovak:


prusnak:
Don't send work in progress stuff, send when there is actually something to review and merge.

PR #153, opened 2018-06-29
protob: add a smart pb2py builder

matejcik:
Now we don't need `build_protobuf` anymore and this is usable by both core and python-trezor (as well as generating custom protobufs from other sources)

We still need protoc, unfortunately, but pb2py now calls it by itself. (little more robustly; instead of generated python classes, it uses the FileDescriptorSet output which is parsable by a built-in protobuf class)

To support the script, messages.proto and types.proto must set a common package. Also there is currently no support for compiling more than one proto file, we depend on the fact that messages.proto import types.proto.
(if this is needed, it should be relatively simple to add, simply pass more than one file to the embedded protoc call)


prusnak:
utACK

Feel free to merge this if this produces the same results as previously
used build_protobuf and send PR to trezor-core if needed.

protob/pb2py
```diff
@@ -0,0 +1,357 @@
...
+                protobuf_module=args.protobuf_module,
+            )
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        module_index_name = os.path.join(tmpdir, "__module_index.py")
+        with open(module_index_name, "w") as module_index:
+            descriptor.write_classes(tmpdir, module_index)
+
+        if args.module_index:
+            os.replace(module_index_name, args.module_index)
```
saleemrashid: Unless I've misread the code, this should fail because `tmpdir` is unlikely to be on the same filesystem as `args.module_index`

protob/pb2py
```diff
@@ -0,0 +1,357 @@
...
+                with open(pathname, "r") as f:
+                    if next(f, None) == AUTO_HEADER:
+                        os.unlink(pathname)
+            except Exception:
+                pass
+
+        for filename in os.listdir(tmpdir):
+            src = os.path.join(tmpdir, filename)
+            dst = os.path.join(args.out_dir, filename)
+            os.replace(src, dst)
```
saleemrashid: See previous comment

protob/pb2py
```diff
@@ -0,0 +1,357 @@
...
+
+    parser = argparse.ArgumentParser()
+    # fmt: off
+    parser.add_argument('proto', help="Protobuf definition file")
+    parser.add_argument('out_dir', help="Directory for generated source code")
+    parser.add_argument('-p', '--package', default="trezor", help="Name of protobuf package to extract. Use '-p all' to process all packages")
+    parser.add_argument('-P', '--protobuf-module', default="protobuf", help="Name of protobuf module")
+    parser.add_argument('-l', '--module-index', help="Generate list of modules")
+    parser.add_argument('--message-type', default="MessageType", help="Name of enum with message IDs")
+    parser.add_argument("--protobuf-default-include", default="/usr/include", help="Location of protobuf's default .proto files.")
```
saleemrashid: Is `-I/usr/include` actually necessary? I've never had any issues without it.

saleemrashid:
The `os.replace` is definitely problematic on my system. You could use the `dir` argument to `TemporaryDir`, or use non-atomic file operations.

Also, see pull request #154 against this branch.

saleemrashid:
This probably needs to be run through `flake8`, and there's a mixture of `%` formatting and `str.format()`

protob/pb2py
matejcik: argh. not as unlikely in practice but you're right, I assumed that `replace` is fs-safe unlike `rename`

shutils it is then

protob/pb2py
matejcik: turns out it is. our protobufs require `descriptor.proto` which lives there by default

matejcik:
@saleemrashid re flake8, we'll actually be disabling some of its warnings in favor of `black` (this code is already compliant)
re formatting mix, well... i don't consider it as much of a problem, but maybe worth unification?

protob/pb2py
saleemrashid: Which operating system do you use? On every system I have used, `/tmp` is a `tmpfs` (in-memory filesystem), so this is failing for me.

protob/pb2py
saleemrashid: Which version of `protoc`? Imports fine without, for me at least.

protob/pb2py
matejcik: Linux Mint, /tmp seems to be part of root. I assume this would be the same on Ubuntu, because why would the downstream change this particular thing.

protob/pb2py
matejcik: libprotoc 2.6.1. That's rather old, perhaps the new versions include this by default; nevertheless, i do actually need it :)

protob/pb2py
saleemrashid: Arch Linux uses `tmpfs` and I'm certain other distributions do so.

PR #154, opened 2018-06-29
Do not use temporary file in protoc invocation

saleemrashid:
`protoc` can write the descriptor set to standard output and remove the need for an intermediate file. Also, remove the unnecessary `-I/usr/include`.

Does not fix the `os.replace` issue.

### DIFF
--- a/protob/pb2py
+++ b/protob/pb2py
@@ -33,6 +33,25 @@ FIELD_TYPES = {
 # fmt: on
 
 
+def protoc(filename):
+    dirname, basename = os.path.split(os.path.abspath(filename))
+
+    serialized = subprocess.check_output(
+        [
+            "protoc",
+            "--include_imports",
+            "--descriptor_set_out=/dev/stdout",
+            os.path.join(".", basename),
+        ],
+
+        cwd=dirname,
+    )
+
+    descriptor_set = descriptor_pb2.FileDescriptorSet.FromString(serialized)
+
+    return descriptor_set
+
+
 def remove_from_start(s, prefix):
     if s.startswith(prefix):
         return s[len(prefix) :]
@@ -67,22 +86,17 @@ def import_statement_from_path(path):
 class Descriptor:
     def __init__(
         self,
-        descriptor_file,
+        descriptor_set,
         message_type="MessageType",
         package=None,
         protobuf_module="protobuf",
     ):
-        # parse descriptor
-        self.descriptor = descriptor_pb2.FileDescriptorSet()
-        with descriptor_file as f:
-            self.descriptor.ParseFromString(f.read())
-
         # find relevant files (this omits `descriptor.proto` from trezor package)
         if package is None:
-            self.files = self.descriptor.file
+            self.files = descriptor_set.file
         else:
             self.files = [
-                file for file in self.descriptor.file if file.package == package
+                file for file in descriptor_set.file if file.package == package
             ]
 
         logging.debug(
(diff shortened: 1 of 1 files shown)


matejcik:
i like the general idea, but this probably won't ever work on windows? unlike the original code

saleemrashid:
Yeah, that's an issue. Cannot think of any way to mitigate on Windows, should I close this PR?

Upstream issue hasn't received much attention https://github.com/google/protobuf/issues/4163

saleemrashid:
Alternatively, you could easily refactor this as a `protoc` plugin. You take a `google.protobuf.compiler.plugin_pb2.CodeGeneratorRequest` on standard input (which contains the `FileDescriptorSet`) and write out a `CodeGeneratorResponse` on standard output (with the filenames and their content)

No need to mess around with temporary files

matejcik:
that's again a very good idea, and again somewhat impractical due to protoc's choices :/

Namely: "plugin" must be a stand-alone executable, at least according to [this](https://stackoverflow.com/questions/42444090/how-to-run-a-protobuf-python-plugin-on-windows).
I'll double-check - if it were possible to pass a .py file, that would allow me to make `pb2py` act both as a plugin and a runner, which would be nice.

But I'm going to go ahead and close this with the conclusion that most likely, avoiding temporary files here is more trouble than it's worth, esp. given that a couple lines later I'm using temporary files again to ensure that the conversion doesn't mess up the existing files.

saleemrashid:
Note that `protoc` plugins would not need temporary files (see https://github.com/trezor/trezor-common/pull/155)

Also, the StackOverflow link you have posted shows that it is possible to wrap it in a batch file for Windows support.

matejcik:
you seem to be missing the part where I'm generating the output into a temporary directory anyway?
This is to avoid a situation where the proto regeneration fails partway through and messes up your sources. Maybe protoc does that by itself, but I'm reluctant to rely on it.

Also, the end goal is to have a single script you run, either as part of python-trezor's `python setup.py prebuild`, or as a make step in trezor-core. Using a .bat wrapper is highly impractical here, you'd push the OS-awareness to all consumers of pb2py.

I mean, maybe you could use pb2py as a runner, and have a separate .bat "plugin" for Windows and .py for Unix-likes? Or build upon the `isatty` detection and output the .bat wrapper if you're on windows? None of this is terribly complicated, again, I like the general approach. But it's a series of small hacks that work around protoc's design choices in order to avoid touching a temporary directory -- which was not a goal in the first place.

ISTM the additional elegance gained by avoiding tmpdirs is fully lost in the protoc workarounds; esp. given that protoc either uses tmpdirs internally anyway, or you're actually losing a feature.

A fully correct solution would be to use a separate, non-protoc proto format parser. But none I've seen seemed more practical than invoking protoc.

saleemrashid:
I see. I found the idea of a `protoc` plugin more flexible than invoking `protoc` internally (for example, with all the discrepancies such as requiring `-I/usr/include`).

Also, the temporary directory an issue, but `protoc` handles this more cleanly (you never write to files directly, you pass file names/contents back to `protoc`)

An interesting approach is [Nanopb's](https://github.com/nanopb/nanopb/blob/master/generator/nanopb_generator.py) which detects automatically whether it is being invoked as a `protoc` plugin or not. Although it still requires you to invoke `protoc` yourself in the latter case.

matejcik:
This is also an [interesting approach] btw :)  If [this](https://stackoverflow.com/a/48943532/222189) worked with protoc, you *could* make a polyglot script for the plugin. It would still need to be called .bat though :/
(and no i won't be adding it)

The ".bat" name thing is really unfortunate! Otherwise you could put the wrapper in `protoc-gen-something.bat` and the py file in `protoc-gen-something` and you'd have the same `protoc --plugin=something` command line. Nanopb seems to have the same problem.

Anyway I took inspiration from this PR, extracted the `protoc()` function (with tempdir; it didn't seem worth it to use the new code on Unixes if i was going to retain the tempdir approach as fallback) and verified that the new code actually does work on Windows. The additional "-I/usr/include" produces a warning, but doesn't interfere otherwise, which is nice too.

[interesting approach]: https://github.com/fireflyoo/polyglot/blob/master/Python.py.bat

matejcik:
FWIW Windows have a "Named Pipes" facility, but it is extremely awkward to work with from Python. I tried to write a bit of code inspired by [this](https://stackoverflow.com/a/50403960/222189) but it fell on my head; and in any case it really doesn't seem worth the hassle.

PR #155, opened 2018-07-01
Refactor pb2py as protoc plugin

saleemrashid:
Needs testing on Windows. Removes need for `tempfile`.

If invoked directly (and standard input is a TTY), it will output usage and examples.


matejcik:
Closing as discussed in #154. This is a good approach that I'll keep in mind, but for now it's less suitable for the task at hand than the original solution.

PR #111, opened 2018-05-23
Change URL for Bitcore (Monacoin).

cryptcoin-junkey:
mona.chainsight.info is not maintained well and it cannot handle SegWit transactions.
Newer URL can handle them.

PR #157, opened 2018-07-03
Change URL for Bitcore (Monacoin).

wakiyamap:
related https://github.com/trezor/trezor-common/pull/111

I am developing a monacoin version of electrum, 
but since there is a possibility that a problem will occur with an old explorer.

I will PullRequest about blockhook this week.

OK https://insight.electrum-mona.org/api/block-index/0


prusnak:
URL added in 97345e83f3b68cfb81f1301b8b2eb53e256f149a

PR #160, opened 2018-07-04
Fix typo for Ethereum Social [ETSC]

akx20000a:
We use Ethereum Social for our coin name instead of EthereumSocial


akx20000a:
@prusnak Wow, it was fast 🚀 , Awesome 👍

PR #162, opened 2018-07-09
Ripple protobuf messages

tsusanka:
Adds Ripple protobuf messages. Other Ripple PRs will follow in trezor-core and trezor-python after this is merged.

### DIFF
--- a/protob/messages-ripple.proto
+++ b/protob/messages-ripple.proto
@@ -1,0 +1,55 @@
+syntax = "proto2";
+
+// Sugar for easier handling in Java
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageRipple";
+
+/**
+ * Request: Address at the specified index
+ * @next RippleAddress
+ */
+message RippleGetAddress {
+	repeated uint32 address_n = 1;			// BIP-32 path. For compatibility with other wallets, must be m/44'/144'/index'
+	optional bool show_display = 2;			// optionally show on display before sending the result
+}
+
+/**
+ * Response: Address for the given index
+ * @prev RippleGetAddress
+ */
+message RippleAddress {
+	optional string address = 1;	// Address in Ripple format (base58 of a pubkey with checksum)
+}
+
+/**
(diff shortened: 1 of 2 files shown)


protob/messages-ripple.proto
```diff
@@ -0,0 +1,63 @@
...
+	optional string address = 1;	// Address in Ripple format (base58 of a pubkey with checksum)
+}
+
+/**
+ * Request: ask device to sign Ripple transaction
+ * @next RippleTxTypeRequest
+ */
+message RippleSignTx {
+	repeated uint32 address_n = 1;				// BIP-32 path. For compatibility with other wallets, must be m/44'/144'/index'
+	optional string account = 2;				// source account address. This address must be equal to the one derived from the BIP-32 path
```
prusnak: Can we drop this? It seems redundant.

tsusanka: I wasn't sure about that either. The reason I left it was that it is a confirmation that the user is sending funds from the account he assumes. It is used only [here](https://github.com/trezor/trezor-core/blob/tsusanka/ripple/src/apps/ripple/sign_tx.py#L18).

But I'm not keen on it, maybe it's a bit far-fetched.

tsusanka:
@prusnak I think/hope this is the final version, so it's ready for review

protob/messages-ripple.proto
```diff
@@ -0,0 +1,55 @@
...
+	optional string address = 1;	// Address in Ripple format (base58 of a pubkey with checksum)
+}
+
+/**
+ * Request: ask device to sign Ripple transaction
+ * @next RippleSignedTx
+ */
+message RippleSignTx {
+	repeated uint32 address_n = 1;				// BIP-32 path. For compatibility with other wallets, must be m/44'/144'/index'
+	optional uint32 fee = 2;					// fee (in drops) for the transaction
```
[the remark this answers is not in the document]
tsusanka: You're correct. Same goes for the payment's `amount`, I guess. I fixed both

prusnak:
LGTM Merged

PR #164, opened 2018-07-11
Monero protocol messages added

ph4r05:
Initial Monero Protocol messages:
 - get watch only credentials
 - key image sync
 - transaction signature

### DIFF
--- a/protob/messages-monero.proto
+++ b/protob/messages-monero.proto
@@ -1,0 +1,412 @@
+syntax = "proto2";
+
+// Sugar for easier handling in Java
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageMonero";
+
+import "messages-common.proto";
+
+
+/**
+ * Structure representing Monero public address
+ */
+message MoneroAccountPublicAddress {
+    optional bytes spend_public_key = 1;
+    optional bytes view_public_key = 2;
+}
+
+/**
+ * Structure representing Monero transaction destination entry
+ */
+message MoneroTransactionDestinationEntry {
+    optional uint64 amount = 1;
+    optional MoneroAccountPublicAddress addr = 2;
+    optional bool is_subaddress = 3;
+}
+
+/**
+ * Structure representing Monero initial transaction information
+ */
+message MoneroTransactionData {
+	optional uint32 version = 1;
+	optional bytes payment_id = 2;
+	optional uint64 unlock_time = 3;
+	repeated MoneroTransactionDestinationEntry outputs = 4;
+	optional MoneroTransactionDestinationEntry change_dts = 5;
+	optional uint32 num_inputs = 6;
+	optional uint32 mixin = 7;
+	optional uint64 fee = 8;
+	optional uint32 account = 9;
+	repeated uint32 minor_indices = 10;
+	optional bool is_multisig = 11;
+	optional bytes exp_tx_prefix_hash = 12;
+	repeated bytes use_tx_keys = 13;
+	optional bool is_bulletproof = 14;
+}
+
+/**
+ * Structure representing Monero list of sub-addresses
+ */
+message MoneroSubAddressIndicesList {
+    optional uint32 account = 1;
+    repeated uint32 minor_indices = 2;
+}
+
+/**
+ * Structure representing Monero UTXO for key image sync
+ */
+message MoneroTransferDetails {
+    optional bytes out_key = 1;
+    optional bytes tx_pub_key = 2;
+    repeated bytes additional_tx_pub_keys = 3;
+    optional uint64 internal_output_index = 4;
+}
+
+/**
+ * Structure representing Monero encrypted exported key image
+ */
+message MoneroExportedKeyImage {
+    optional bytes iv = 1;
+    optional bytes tag = 2;
+    optional bytes blob = 3;
+}
+
+/**
+ * Request: Ask device for public address derived from seed and address_n
+ * @next MoneroAddressAck
+ */
+message MoneroGetAddressRequest {
+    repeated uint32 address_n = 1;			// BIP-32 path to derive the key from master node
+	optional bool show_display = 2;			// Optionally show on display before sending the result
+	optional uint32 network_type = 3;		// Main-net / testnet / stagenet
+	optional uint32 account = 4;            // Major subaddr index
+	optional uint32 minor = 5;              // Minor subaddr index
+}
+
+/**
+ * Response: Contains Monero watch-only credentials derived from device private seed
+ * @prev MoneroGetAddressRequest
+ */
+message MoneroAddressAck {
+    optional bytes address = 1;
+}
+
+/**
+ * Request: Ask device for watch only credentials
+ * @next MoneroWatchKeyAck
+ */
+message MoneroGetWatchKeyRequest {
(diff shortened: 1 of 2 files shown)


protob/messages-monero.proto
```diff
@@ -0,0 +1,438 @@
...
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageMonero";
+
+import "common.proto";
+
+
+/**
+ * Structure representing Monero public address
+ */
+message MoneroAccountPublicAddress {
```
prusnak: Let's drop the `m_` prefix.

ph4r05: Yep why not. I took the field naming from the original Monero code but I don't like it either.

protob/messages-monero.proto
```diff
@@ -0,0 +1,438 @@
...
+}
+
+/**
+ * Structure representing Monero UTXO for key image sync
+ */
+message MoneroTransferDetails {
+    optional bytes out_key = 1;
+    optional bytes tx_pub_key = 2;
+    repeated bytes additional_tx_pub_keys = 3;
+    optional uint64 m_internal_output_index = 4;
```
prusnak: Let's drop the `m_` prefix.

protob/messages-monero.proto
```diff
@@ -0,0 +1,438 @@
...
+    optional MoneroKeyImageSyncStep step = 2;
+    optional MoneroKeyImageSyncFinal final_msg = 3;
+}
+
+/**
+ * Request: Universal Monero protocol implementation diagnosis request.
+ * @next MoneroDiagResp
+ * @next MoneroRespError
+ */
+message MoneroDiag {
```
prusnak: Do we need this in production code? If not, I would rename these messages to contain `Debug` prefix to make it obvious they don't belong to production.

ph4r05: I needed to discuss this. It is not needed for the production code. 
So prefixing with Debug and leaving it in the `messages-monero.proto` is OK?

prusnak: The question is whether we still need these messages or if they were only useful for development and are no longer necessary once the code has been written. If they are still needed, let's use the Debug prefix and keep them in `messages-monero` proto file.

prusnak:
What's the difference between `Tsx` and `Tx` in message names?

ph4r05:
`Tsx` vs `Tx`, I guess it is kind of mixed from multiple sources, but should mean the same, Transaction. Would you prefer unification to `Tx`?

prusnak:
Let's unify to `Transaction`. There is no reason to use short names in message/field names.

I would also expand `Addr` -> `Address`, `Rct` -> `RingCt`. The more decriptive names, the better.

Also I don't understand the `MoneroTsxMlsagDone` name.

ph4r05:
The Mlsag stands for `Multilayered Spontaneous Anonymous Group Signatures`, the protocol step corresponds to the 
https://github.com/monero-project/monero/blob/a844844cda9a0beb62d95015f92327366d73b3a1/src/ringct/rctSigs.cpp#L715

The long message name version would thus be `MoneroTransactionMultilayeredSpontaneousAnonymousGroupSignaturesDone` or better `MoneroTransactionPreMultilayeredSpontaneousAnonymousGroupSignaturesHashDone`.

ph4r05:
I think I will need some benchmarking w.r.t. abbreviations expansions. I am not sure whether to expand `MLSAG` either. `Rct` -> `RingCt` makes sense, `RingConfidentialTransaction` would be a bit messy. 

The question is whether aforementioned message should be named `MoneroTransactionPreMultilayeredSpontaneousAnonymousGroupSignaturesHashDone` and how long message identifiers can `protoc` handle. I would prefer `mlsag` abbreviation.

Some field names are taken directly from the Monero serialization schemes, e.g., 
https://github.com/ph4r05/monero-serialize/blob/1a582fbfdf0990c5f9155897207d71188373383e/monero_serialize/xmrtypes.py#L795

The thing is I would like to preserve message field names for these serialization schemes as data could be serialized also in JSON and having different field names could ruin the deserialization and cause pain later. 

This is the case for RPC JSON and KV serialization used e.g., in the wallet files. These cases affect monero-glue code, not trezor-core.monero but the question is whether we want to have different field names for same messages here.

prusnak:
I agree that we don't want to expand everything and it makes sense to leave the field names as is - to reflect their counterparts in Monero code.

I propose the following changes:

a) change `Tx` and `Txs` to `Transaction` in message names
b) change `MoneroSubAddrIndicesList` to `MoneroSubAddressIndicesList`

To discuss:

c) can we drop `MoneroRespError` completely? - we don't have specific Error messages for any other coins and we don't want to leak exceptions from Trezor anyway (as they might contain sensitive info) - so we usually return `Failure(FirmwareError)` message in most of the cases

d) most of other coins follow the `FooRequest`/`FooAck` scheme, while you use `Foo`/`FooResp` scheme - I agree that the original scheme is not ideal, that's why I am not pushing for it too hard, but maybe we can use that for consistency

What do you think?

ph4r05:
Yep makes sense! I am on it. 

d) definitely, in production there should not be stack trace. I can remove this message completely.

e) I can switch it to `FooRequest`/`FooAck`, no problem.

ph4r05:
refactoring finished

prusnak:
I merged the PR, thanks!

I did some refactoring after the merge too (in a5e6dff3ff5ff534bf0b98e977c44ec4e9341d4a ):

1) I used nested messages, we haven't use this before, but it really unclutters the definition by hiding definitions used in exactly one place into the message itself.

2) I dropped the `@prev` directive from the comments, it's superfluous.

I don't think we need `MoneroTransactionSignRequest` and `MoneroKeyImageSyncRequest`. Why not use the messages they wrap directly? If you I agree, I'll just drop them from the definition.

ph4r05:
Thanks for the merge!

Regarding the wrapping messages - I find it more clear to have the whole
protocol encapsulated in the particular wrapping message. There is only one
message handler registered for the particular purpose (tx sign, ki sync)
and the protocol sub-message multiplexing is hidden in the protocol
handler. For tx sign I would need to register 7 different request handlers,
route it to the single tx sign handler and do the message multiplexing
again. I found this approach more messy.

The protocols for tx sign and ki sync are stateful, so it also made sense
to tie the messages together in the wrapping request message and hide
undrelying multistep logic in ththe handler.
Moreover it is easy to extend the protocol with the common request fields
if needed in the future.

Regarding the embedded messages - I will check how the imports work in the
trezor-core code. If that works well after the changes. But I guess the
embedded messages are still OK to import.
ne 22. 7. 2018 v 20:07 odesílatel Pavol Rusnak <notifications@github.com>
napsal:

prusnak:
> Regarding the embedded messages - I will check how the imports work in the trezor-core code.

We did the same trick for other messages and currently, there is no difference in trezor-core code whether the messages are nested or not.

> Maybe the wrapping messages will make more sense after I make PR to trezor-core so the benefit is visible from the code.

OK

ph4r05:
@prusnak I've just notices a little bug in the last Monero refactoring. 
`MoneroGetAddressRequest` got renamed to `MoneroGetAddress` but the wire type in `messages.proto` is not changed so the request does not have a wire number. 

The same holds for other messages from the refactoring batch.

PR #166, opened 2018-07-13
 Add Tron support

fbsobreira:


prusnak:
Can you please squash your commits into one and rebase on top of current master?

benjaminchodroff:
This support for Tron in Trezor looks great. I would love to see this added to Trezor soon so we can use a hardware token for multi-signature wallet security.

fbsobreira:
Thanks for the feedback. I have also rebase to the last commit!

nesttle:
Nice work! Hope we can see this implemented soon in Trezor One.

protob/messages-tron.proto
```diff
@@ -0,0 +1,207 @@
...
+    optional string address = 1;    // Tron address
+}
+
+/**
+ * Request: Ask device to sign Tron transaction
+ * @next TronSignedTx
+ */
+message TronSignTx {
+    repeated uint32 address_n = 1;              // BIP-32 path to derive the key from master node
+    optional TronTransaction transaction = 2;   // Transaction data
```
matejcik: is there any reason this is a separate message, as opposed to having the fields inside `TronSignTx` directly?

protob/messages-tron.proto
```diff
@@ -0,0 +1,207 @@
...
+        // Update Asset
+        message TronUpdateAssetContract {
+            optional bytes owner_address = 1;               // Account address
+            optional bytes description = 2;                 // New description
+            optional bytes url = 3;                         // New URL
+        }
+        // Network proposal contract
+        message TronProposalCreateContract {
+            optional bytes owner_address = 1;               // Account address
+            map<uint64, uint64> parameters = 2;             // Parameter to be changed
```
matejcik: we currently don't support implicit maps, and we probably don't want to; please convert to explicit repeated "(name, value)" submessage.

protob/messages-tron.proto
```diff
@@ -0,0 +1,207 @@
...
+        optional TronProposalDeleteContract proposal_delete_contract = 18;
+        optional TronSetAccountIdContract set_account_id_contract = 19;   
+    }
+    // Common part of transaction
+    message TronTransaction {
+        optional bytes ref_block_bytes = 1;     // reference block number 
+        optional bytes ref_block_hash = 4;      // reference block hash
+        optional uint64 expiration = 8;         // transaction expiration
+        optional bytes data = 10;               // Extra transaction info
+        repeated TronContract contract = 11;    // contract messages
```
matejcik: if this is repeated, it should not be part of a single big `TronSignTx` message. Instead, make a `TronTxRequest` and `TronTxAck` that replies with the individual contract messages - see [this comment](https://github.com/trezor/trezor-common/pull/197#issuecomment-416635279) on EOS for the recommended flow.

matejcik:
and my favorite question: why does every `TronContract` message repeat a "from address", if you have a `GetAddress` message? Isn't the "from address" implicit from the derived `address_n` ?

protob/messages-tron.proto
fbsobreira: I have tried to keep it as close as we have in our protocol structure to facilitate the maintenance. Have it updated now.

protob/messages-tron.proto
fbsobreira: OK, updated.

protob/messages-tron.proto
fbsobreira: No need to be repeated, have changed to optional.

fbsobreira:
Thanks @matejcik ! About the address, I will be deriving from path... I was trying to make the code as close as we have in our protocol to facilitated maintenance. 
Please let me know if there is anything else I should address

matejcik:
what does the `data` ("extra transaction info") field contain?

if you structured this according to your protocol, does that mean that now `TronSignTx` matches the original `TronTransaction`, except for the `address_n` field and the addresses in the individual contract messages?

What is the format of the messages in your protocol / API? Will it be possible to have a direct conversion, e.g., from JSON, or maybe from XDR or other standardized format? Or do you intend to integrate Trezor support into your software directly?

fbsobreira:
`data` is generic field where users can add text to transactions, like reference messages. 

Its similar, I had also changed other messages. But everything is ok, with this info I can get all fields needed and pack the correct message in trezor-core.

We use protobuf3. Dont think we can do a direct conversion. We will have Trezor integrated into our wallet with Trezor connect.

protob/messages.proto
```diff
@@ -159,6 +159,14 @@ enum MessageType {
     MessageType_StellarBumpSequenceOp = 221 [(wire_in) = true];
     MessageType_StellarSignedTx = 230 [(wire_out) = true];
 
+    // TRON
+	MessageType_TronGetAddress = 250 [(wire_in) = true];
```
matejcik: please fix the indent here

matejcik:
If `data` is for user messages, shouldn't it be a `string`? Will that field be processed? Will it be signed? Will it be displayed on Trezor screen?

protob/messages.proto
fbsobreira: done!

fbsobreira:
`data` changed to string. It is signed too, but optional, and will be displayed on Trezor screen?

matejcik:
Yes, if it is signed, it should be displayed on screen. OK then.
Looking over it once more, it seems you have most `string` fields set as `bytes` - URL, description, etc.

How about addresses, are those supposed to be strings or bytes? Which representation is signed, decoded (from base58 AFAICT?) bytes, or the base58 string?
(don't change this just yet either way, we now have an internal discussion about which way we want this to go)

Do you use the public key for anything? On some coins, it is useful to be able to verify signed messages - but you don't have a sign/verify functionality. On other coins it's possible to export something akin to Bitcoin's xpub and derive child addresses. If that's your case, you should rename the method to `GetXpub` or `GetPublicNode`. Otherwise please remove the method - we don't need it on every coin, if the public key is not actually used for anything.

fbsobreira:
Im reviewing all byte fields and changing to string, where it fits.

Address we sign the decoded from base58, but I think would be better to send address string (base58) and decode on Trezor. Let me know which way is better for Trezor.

We don't have a use case for public key at the moment. I will remove it for now, and if needed we can request another PR later, correct?

I will be adding a serialized_tx to TronSignedTx message, just like in ripple. Is that ok?

I will be waiting for the info about address to update the PR! thank you for your time!

matejcik:
Thanks.

Re addresses, let's send them already decoded, i.e., no change from what's in the protobufs now.

One more thing confuses me: `ref_block_bytes` says it's a number in the comment, but its type (and name) is `bytes`. What does this field actually contain?

Adding `serialized_tx` is OK.
You can also add the public key methods in a later PR if it becomes necessary.

Mostly LGTM now, waiting for your update with strings where appropriate.

fbsobreira:
Hi @matejcik !!! Any updates or suggestions??? We have the trezor-core and python-trezor already... Should we wait for this PR to be merged or can we submit for evaluation now?? Thank you

matejcik:
Thanks for the ping. The PR looks good now. Will probably merge tomorrow (@prusnak, want to look at it first?).

Please submit PRs to core and python-trezor. We will ask you to add a commit updating the trezor-common reference after this PR is accepted, but we can evaluate them as it is now.

prusnak:
utACK for the merge

PR #170, opened 2018-07-16
Tezos: field updates

Adman:
* added curve type option
* returning public key as a b58 encoded string with prefix

@prusnak According to http://doc.tzalpha.net/api/p2p.html#public-key-determined-from-data-8-bit-tag Tezos supports P256 curve. @matejcik told me that trezor supports this curve, but I found only `nist256p1`.  Are they the same?

### DIFF
--- a/protob/messages-tezos.proto
+++ b/protob/messages-tezos.proto
@@ -5,6 +5,15 @@ option java_package = "com.satoshilabs.trezor.lib.protobuf";
 option java_outer_classname = "TrezorMessageTezos";
 
 /**
+ * Tezos Curve Type
+ */
+enum TezosCurveType {
+    Ed25519 = 0;
+    Secp256k1 = 1;
+    P256 = 2;
+}
+
+/**
  * Request: Ask device for Tezos address corresponding to address_n path
  * @start
  * @next TezosAddress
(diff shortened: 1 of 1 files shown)


Adman:
Rebased, should be ready for merging.

I removed `TezosOperationCommon` field and copied its all subfields to each operation. The reason is that I had to add `Reveal` operation, which is for revealing the public key to the network (and is called when the account is doing first transaction/origination/delegation in a lifetime). This operation may be bundled with transaction/origination/delegation, so one has to fill out common fields in reveal and also in transaction...

Also, as I mentioned in my first comment, I added 3 different curve types.

prusnak:
`nist256p1` is NIST P-256, they are the same.

However, I have a question - is there a reason why Trezor should support different curve than just the Ed25519? What are use cases for using others? Would it hurt if we supported just the only one curve? I don't want to introduce unnecessary complexity for no good reason.

Adman:
According to [docs](http://tezos.gitlab.io/betanet/api/p2p.html#public-key-determined-from-data-8-bit-tag) Tezos supports 3 curves. To determine which curve is account using, Addresses/public keys/signatures are distinguishable by prefixes `tz1, tz2, tz3` / `edpk, sppk, p2pk` / `edsig, spsig1, p2sig` according to curve used. Tezos developers added `p256` (tz3) because they wanted to support iphone as a hardware wallet. It would be great if we could support all of them within Trezor.

prusnak:
Thank you!

PR #173, opened 2018-07-20
unified coin data management

matejcik:
This is a preliminary PR for review purposes. It still needs tools to generate code with the new unified method. That is unfinished.

I also still intend to clean up the rest of `coins_details.py` a little.

---

Esp. please review the regenerated `coins_details.json` - a nicely diffable version is here: https://github.com/trezor/trezor-common/commit/a8a7ae0dd985e4873ae3fd8ee15d98b6d5310db7
I'm generating the file from scratch, incl. stuff like Monero, Lisk etc (see `others.json`)
That's why a couple new things like Callisto, EtherSocial, GoChain showed up. They were disabled previously, but now I read them from `defs/ethereum/networks.json`. Should they be disabled?

In addition, I'm hiding every coin that has "Testnet" in its name, is that OK? 
Other things came unhidden, e.g. Viacoin, EOS erc20 (maybe that should be hidden because of EOS mainnet?)

A bunch of erc20 tokens was dropped because they have conflicting symbols. This is an upstream problem but we discussed this with @prusnak and decided to drop support for them altogether.

About 3 tokens also lost their URLs and thus became hidden, because their servers are dead.
(URLs are now taken purely from the upstream token repo, so if we want to readd them, we'd have to do a PR. I already did that for some 64 URLs and the PR was accepted.)


prusnak:
> Should they be disabled?

No, it's fine.

> In addition, I'm hiding every coin that has "Testnet" in its name, is that OK?

Yes. Let's also hide ETH networks with "Testnet" in their name.

defs/support.json
```diff
@@ -1,100 +1,245 @@
...
-		"Dash": "https://electrum.dash.org",
-		"Groestlcoin": "https://www.groestlcoin.org/groestlcoin-electrum-wallet/",
-		"Koto": "https://electrum.kotocoin.info",
-		"Litecoin": "https://electrum-ltc.org",
-		"Monacoin": "https://electrum-mona.org",
-		"Namecoin": "https://github.com/namecoin/electrum-nmc",
-		"Viacoin": "https://vialectrum.org",
-		"Zcoin": "https://github.com/zcoinofficial/electrum-xzc"
-	}
+    "btc:BCH": {
```
prusnak: Maybe use prefix `coin:` instead of `btc`? I understand why you chose that, but I think `coin` is less confusing.

defs/support.json
```diff
@@ -1,100 +1,245 @@
...
+        "trezor2": "2.0.5",
+        "connect": true,
+        "webwallet": true,
+        "other": {
+            "Dash Electrum": "https://electrum.dash.org"
+        }
+    },
+    "btc:DCR": {
+        "connect": null,
+        "electrum": null,
```
prusnak: Does it make sense to have `"electrum": null,`, when we put Electrum links in `other` when defined?

defs/support.json
```diff
@@ -1,100 +1,245 @@
...
+        "connect": null,
+        "webwallet": null
+    },
+    "network:ADA": {
+        "trezor1": null,
+        "trezor2": "soon",
+        "connect": null,
+        "webwallet": null
+    },
+    "network:XMR": {
```
prusnak: Not sure about the `network` prefix, but as I cannot find a better word, let's keep it for now I guess.

matejcik:
this is now ready for review.

It should probably be merged in lockstep with trezor-mcu and trezor-core PRs, which are not ready yet.

Esp. @slush0 please have a look at the regenerated `coins_details.json`. This commit is intended for easy comparison: https://github.com/trezor/trezor-common/pull/173/commits/288445cecc9e18115aa0e46cb6debfb6f8a13c76 -- it uses the same coinmarketcap data and same keys.
Later commits change key names, so coins are shuffled around, and update market caps, so numbers are different throughout the file.

prusnak:
I am OK with the changes

matejcik:
Just noticed that as written, gen_coins.py requires python 3.6.

I'd keep it (we all have py3.6), but something to be aware of. It's also just one place that I know of, so fixing this for lower Pythons wouldn't be a problem, if desired.

prusnak:
utACK, you can merge this!

PR #174, opened 2018-07-22
make signature_v 64bit to support full 32bit chain_id

hackmod:


mkrufky:
Very nice.  This, in conjunction with https://github.com/trezor/trezor-mcu/pull/385 should help immensely with PIRL support :+1:

hackmod:
sorry for inconvince.

should check more testsuite soon.

prusnak:
Sorry, I am against this change. I think that 2 billions chain_ids are enough for everyone and using higher values than 2147483630 might break lots of other software.

PR #175, opened 2018-07-23
Monero: wire type fix

ph4r05:
MoneroGetAddressRequest got renamed to MoneroGetAddress but the message name for wire type in messages.proto is not changed so the request does not have a wire number.


prusnak:
Sorry :-/ Merged

PR #180, opened 2018-07-27
Monero def added

ph4r05:
An attempt to add Monero definitions.

I am not sure how necessary is this and some fields may be completely wrong.


prusnak:
`defs/coins/` directory is only for Bitcoin-like altcoins.

Thanks for the info anyway, some values might be useful later. @matejcik is working on creating a structure for non-Bitcoin-like coins and we'll cherry-pick, the values from this json there.

That's why I'll keep this open, but we will not merge this directly.

ph4r05:
Yep great, thas what I thought. At least there is monero.png :) 
The values would need double check also.

prusnak:
curve and github links added in d72b787d857fdd1bd03c2a45b863b5904077dd54

e don't need/use any other parameters at the moment

PR #184, opened 2018-08-01
Add Ontology support

backslash47:
Hi,

I prepared protobuff messages for Ontology support. Currently there are 5 methods defined:
getPublicKey, getAddress, signTx, signTransfer and signWithdrawOng

This will provide support for all necessary wallet functionality.


backslash47:
Hello, I don't want to be rude, but is there a timeframe for reviewing this PR?

prusnak:
End of next week. I am on a sick leave at the moment.

backslash47:
@prusnak Thanks for the response. Hope you get well soon.

protob/messages-ontology.proto
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Request: Ask device to sign Ontology transfer
+ * @next OntologySignedTransfer
+ */
+ message OntologySignTransfer {
+    repeated uint32 address_n = 1;                  // BIP-32 path to derive the key from master node
+    optional OntologyTransaction transaction = 2;
+    optional OntologyTransfer transfer = 3;
+    optional bool return_payload = 4;
```
prusnak: We don't do this. Remove this field and always return the payload.

protob/messages-ontology.proto
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Request: Ask device to sign Ontology ONG withdrawal
+ * @next OntologySignedWithdrawOng
+ */
+ message OntologySignWithdrawOng {
+    repeated uint32 address_n = 1;                  // BIP-32 path to derive the key from master node
+    optional OntologyTransaction transaction = 2;
+    optional OntologyWithdrawOng withdrawOng = 3;
+    optional bool return_payload = 4;
```
prusnak: We don't do this. Remove this field and always return the payload.

protob/messages-ontology.proto
```diff
@@ -0,0 +1,147 @@
...
+message OntologySignedTransfer {
+    optional bytes signature = 1;
+    optional bytes payload = 2;
+}
+
+/**
+ * Request: Ask device to sign Ontology ONG withdrawal
+ * @next OntologySignedWithdrawOng
+ */
+ message OntologySignWithdrawOng {
```
prusnak: Why do we need separate message for WithdrawOng? Isn't SignedTransfer with asset==ONG the same thing?

protob/messages-ontology.proto
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * Request: Ask device to sign Ontology arbitrary transaction
+ * @next OntologySignedTx
+ */
+ message OntologySignTx {
+    repeated uint32 address_n = 1;                  // BIP-32 path to derive the key from master node
+    optional OntologyTransaction transaction = 2;
+    optional bytes payload = 3;
```
prusnak: We don't allow signing unknown scripts. I suggest to drop this message completely.

protob/messages-ontology.proto
backslash47: No it is not. Those are two different native contract calls. Also the transaction will look different.

protob/messages-ontology.proto
backslash47: The payload is the script built from the transfer object. Client can build it himself. Only he needs to be sure that the result payload is the same as the one built in trezor code.

The payload is not the signature. The signature is returned always. But I don't have problem removing this option and returning the built script always.

protob/messages-ontology.proto
backslash47: That's reasonable. There a few more types of transaction I need to add besides transfer and withdrawOng. Mainly registering Identity on Ontology network (that's no problem) and calling arbitrary smart contract. 

Will you accept signing such call? The script will be build from Smart contract address, smart contract method and arbitrary parameters. I can not guaranty what will be the outcome of calling such contract.

prusnak: Let's create a message for registering an Identity.

I am not so sure about SmartContractCall message, because we cannot show arbitrary parameters in a meaningful way on Trezor. So I would leave that out for the time being.

PR #185, opened 2018-08-03
adjust smartcash info

leoreinaux:


matejcik:
in addition to possibly bad curve name (https://github.com/trezor/trezor-crypto/pull/170), you should use underscore in the definition, i.e. `secp256k1_sha3`

prusnak:
No response, closing.

PR #190, opened 2018-08-08
Lowering dash minfee

karel-3d:
Apparently Dash itself returns fee 220

I am leaving this as 1000 as wallet might have more issues with fee < 1sat/B (fee here is satoshi/kB, daemon returns BTC/kB on fee estimation, gui has sat/B)


prusnak:
ACK, please merge.

PR #192, opened 2018-08-14
XMR: protob protocol improvements

ph4r05:
- Uses only protobuff messages for the transaction construction data - forward compatible. Original transaction source entry and related messages are not stable on Monero side, serialization format of the cryptonote serialization could be changed in the future (new fields), CN serialization does not support versioning, thus it is not forward compatible.
- MoneroTransactionDestinationEntry, MoneroTransactionSourceEntry have to be defined in the root level as they are used in various protocol messages

### DIFF
--- a/protob/messages-monero.proto
+++ b/protob/messages-monero.proto
@@ -48,6 +48,23 @@ message MoneroWatchKey {
 }
 
 /**
+ * Structure representing Monero transaction destination entry
+ */
+message MoneroTransactionDestinationEntry {
+	optional uint64 amount = 1;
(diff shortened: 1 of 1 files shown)


prusnak:
LGTM!

It seems there are two more structures represented as `bytes`: `xmrtypes.TxinToKey` and `xmrtypes.TxOut`. Are we OK with that?

ph4r05:
Thanks for the review!

Ad `xmrtypes.TxinToKey` and `xmrtypes.TxOut` - those are fine as they are directly part of the transaction that goes to the block chain, the format is not going to change significantly and if yes we need also some refactoring.

PR #196, opened 2018-08-23
Monero: support for lite protocol added

ph4r05:
I've implemented the very simple fallback protocol for the Monero wallet (lite protocol / proxy protocol) which is already implemented in the monero C++ codebase.

Should there be some problem with the merge of the high level protocol this could be nice fall back.


prusnak:
Let's leave this unmerged for now. If the high-level proto is rejected, we can revisit this.

prusnak:
I guess we don't need this anymore. Thx!

PR #197, opened 2018-08-26
EOS integration

tarassh:
This pull request adds basic EOS messages to trezor.

`EosGetAddress` Ask device for Eos address corresponding. 
`EosAddress` Contains an Eos address derived from device private seed.

`EosSignTx` Ask device to sign transaction.
`EosTxRequest` Device asks for more data from transaction payload, or returns the signature.
`EosTxAck` Transaction payload data.

`EosTxHeader` represents EOS transaction header

### DIFF
--- a/protob/messages-eos.proto
+++ b/protob/messages-eos.proto
@@ -25,51 +25,120 @@ message EosGetAddress {
 }
 
 /**
+* Structure representing EOS transaction header
+*/
+message EosTxHeader {
+    required uint32 expiration = 1;           // time at which transaction expires
+    required uint32 ref_block_num = 2;        // 16-bit specifies a block num in the last 2^16 blocks.
+    required uint32 ref_block_prefix = 3;     // specifies the lower 32 bits of the blockid at get_ref_blocknum
+    required uint32 max_net_usage_words = 4;  // upper limit on total network bandwidth (in 8 byte words) billed for this transaction
+    required uint32 max_cpu_usage_ms = 5;     // 8-bit upper limit on the total CPU time billed for this transaction
+    required uint32 delay_sec = 6;            // number of seconds to delay this transaction for during which it may be canceled.
+}
+
+/**
+* Structure representing action permission level
+*/
+message EosPermissionLevel {
+    required string actor = 1;
+    required string permission = 2;
+}
+
+/**
+* Structure representing the common part of every action
+*/
+message EosActionCommon {
+    required string account = 1;                    // contract name
+    required string name = 2;                       // action name
+    repeated EosPermissionLevel authorization = 3;
+}
+
+message EosContextFreeAction {
+    required EosActionCommon common = 1;
+    required bytes data = 2;
+}
+
+message EosActionTransfer {
+    required EosActionCommon common = 1;
+    required string from = 2;
+    required string to = 3;
+    required string asset = 4;                      // asset format '1.0000 EOS'
+    optional string memo = 5;
+}
+
+message EosActionDelegate {
+    required EosActionCommon common = 1;
+    required string from = 2;
+    required string receiver = 3;
+    required string net_quantity = 4;               // asset format '1.0000 EOS'
+    required string cpu_quantity = 5;               // asset format '1.0000 EOS'
+    required bool transfer = 6;                     // transfer delegated tokens or not.
+}
+
+message EosActionUndelegate {
+    required EosActionCommon common = 1;
+    required string from = 2;
+    required string receiver = 3;
+    required string net_quantity = 4;               // asset format '1.0000 EOS'
+    required string cpu_quantity = 5;               // asset format '1.0000 EOS'
+}
+
+message EosActionBuyRam {
+    required EosActionCommon common = 1;            
+    required string payer = 2;
+    required string receiver = 3;
+    required string quantity = 4;                   // asset format '1.0000 EOS'
+}
+
+message EosActionBuyRamBytes {
+    required EosActionCommon common = 1;
+    required string payer = 2;
+    required string receiver = 3;
+    required uint32 bytes = 4;
+}
+
+message EosActionSellRam {
+    required EosActionCommon common = 1;
+    required string account = 2;
+    required int64 bytes = 3;
+}
+
+message EosActionVoteProducer {
+    required EosActionCommon common = 1;
+    required string voter = 2;                      // voter account
+    optional string proxy = 3;                      // proxy voter account
+    repeated string producers = 4;                   // list of producers
+}
+
+/**
  * Request: Ask device to sign transaction
- * All fields are optional from the protocol's point of view. Each field defaults to value `0` if missing.
- * Note: the first at most 1024 bytes of data MUST be transmitted as part of this message.
  * @start
  * @next EosTxRequest
  * @next Failure
  */
  message EosSignTx {
-    repeated uint32 address_n = 1;                // BIP-32 path to derive the key from master node 44'/194'/0'
-    optional bytes chain_id = 2;                  // 256-bit long chain id
-    optional EosTxHeader header = 3;              // Eos transaction header
-    optional bytes data_initial_chunk = 4;        // The initial data chunk (<= 1024 bytes)
-    optional uint32 data_length = 5;              // Length of transaction payload
-    
-    /**
-    * Structure representing EOS transaction header
-    */
-    message EosTxHeader {
-        optional uint32 expiration = 1;           // time at which transaction expires
-        optional uint32 ref_block_num = 2;        // 16-bit specifies a block num in the last 2^16 blocks.
-        optional uint32 ref_block_prefix = 3;     // specifies the lower 32 bits of the blockid at get_ref_blocknum
-        optional uint32 max_net_usage_words = 4;  // upper limit on total network bandwidth (in 8 byte words) billed for this transaction
-        optional uint32 max_cpu_usage_ms = 5;     // 8-bit upper limit on the total CPU time billed for this transaction
-        optional uint32 delay_sec = 6;            // number of seconds to delay this transaction for during which it may be canceled.
-    }
+    required uint32 address_n = 1;                // BIP-32 path to derive the key from master node 44'/194'/0'
+    required bytes chain_id = 2;                  // 256-bit long chain id
+    required EosTxHeader header = 3;              // Eos transaction header
+    repeated EosContextFreeAction context_free_actions = 4;        // should be 0
+    optional EosActionTransfer transfer = 5;
+    optional EosActionDelegate delegate = 6;
+    optional EosActionUndelegate undelegate = 7;
+    optional EosActionBuyRam buy_ram = 8;
+    optional EosActionBuyRamBytes buy_ram_bytes = 9;
+    optional EosActionSellRam sell_ram = 10;
+    optional EosActionVoteProducer vote_producer = 11;
+    optional bytes transaction_extensions = 12;
+    optional bytes context_free_data = 13;
 }
 
 /**
- * Response: Device asks for more data from transaction payload, or returns the signature.
- * If data_length is set, device awaits that many more bytes of payload.
- * Otherwise, the signature_* fields contain the computed transaction signature. All three fields will be present.
+ * Response: Device returns the signature.
+ * The signature_* fields contain the computed transaction signature. All three fields will be present.
  * @end
- * @next EosTxAck
  */
-message EosTxRequest {
-    optional uint32 data_length = 1;    // Number of bytes being requested (<= 1024)
+message EosSignTxResponse {
     optional uint32 signature_v = 2;    // Computed signature (recovery parameter, limited to 31 or 32)
     optional bytes signature_r = 3;     // Computed signature R component (256 bit)
     optional bytes signature_s = 4;     // Computed signature S component (256 bit)
-}
-
-/**
- * Request: Transaction payload data.
- * @next EosTxRequest
- */
- message EosTxAck {
-    optional bytes data_chunk = 1;  // Bytes from transaction payload (<= 1024 bytes)
 }
--- a/protob/messages-eos.proto
+++ b/protob/messages-eos.proto
@@ -1,0 +1,75 @@
+syntax = "proto2";
+package hw.trezor.messages.eos;
+
+// Sugar for easier handling in Java
+option java_package = "com.satoshilabs.trezor.lib.protobuf";
+option java_outer_classname = "TrezorMessageEos";
+
+/**
+ * Request: Ask device for Eos address corresponding to address_n path
+ * @start
+ * @next EosAddress
+ * @next Failure
+ */
+message EosGetAddress {
+    repeated uint32 address_n = 1;  // BIP-32 path to derive the key from master node 44'/194'/0'
+    optional bool show_display = 2; // optionally show on display before sending the result
+}
+
+/**
+ * Response: Contains an Eos address derived from device private seed
+ * @end
+ */
+ message EosAddress {
+    required string address = 1;     // EOS address in Base58 encoding
+}
+
+/**
+ * Request: Ask device to sign transaction
+ * All fields are optional from the protocol's point of view. Each field defaults to value `0` if missing.
+ * Note: the first at most 1024 bytes of data MUST be transmitted as part of this message.
+ * @start
+ * @next EosTxRequest
+ * @next Failure
+ */
+ message EosSignTx {
+    repeated uint32 address_n = 1;                // BIP-32 path to derive the key from master node 44'/194'/0'
+    optional bytes chain_id = 2;                  // 256-bit long chain id
+    optional EosTxHeader header = 3;              // Eos transaction header
+    optional bytes data_initial_chunk = 4;        // The initial data chunk (<= 1024 bytes)
+    optional uint32 data_length = 5;              // Length of transaction payload
+    
+    /**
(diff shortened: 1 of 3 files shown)


prusnak:
Trezor will not sign arbitrary data without understanding them and
showing them on display.
You should split `data` into reasonable messages, so thr fields can be
parsed and shown on display.
Currently this is not possible.

prusnak:
Maybe just add message similar to SteemOperationalTransfer along
EosTxHeader:
https://github.com/trezor/trezor-common/commit/cd1a963f3189a7318e40daee27dab336ab66d50f#L819-L825

tarassh:
> Trezor will not sign arbitrary data without understanding them and
> showing them on display.
> You should split `data` into reasonable messages, so thr fields can be
> parsed and shown on display.
> Currently this is not possible.


The data will be parsed and displayed help of internal logic. If the data will not fit the decoding scheme it with throw an error. The reason is that in EOS every transaction communicates with smart contracts.
There are at least  7 actions that need to be implemented in order to have valid integration into EOS ecosystem, soon could be more:

- transfer - this is the action(method) of `eosio.token` smart contract and other compatible token contracts.
- delegate - all of the actions below are parts of `eosio` smart contract
- undelegate
- buy ram
- buy ram bytes
- sell ram
- vote produce

The intent to have generic interface which will not change over time, but internal logic only.

prusnak:
You need to implement these 7 actions via messages. Generic interface is not acceptable as this introduces parsing of the user data in the firmware, which is unwanted.  I suggest you look at messages of other coins to get the idea how these messages could look.

tarassh:
Got it.

Just for your information. Steem transfer like message, provided below, and other messages anyway will be converted into format acceptable for signing. Fields `from`, `to`, `amount`,` asset` all this should be represented in uint64 before signing.

```
/**
  * Steem related message
  */
 message SteemOperationTransfer {
 	required string from = 1;
 	required string to = 2;
 	required int64 amount = 3;
 	required string asset = 4;
 	optional string memo = 5;
 }
```

prusnak:
Are you use that `from` and `to` are uint64 not uint256? Always use the type that is closest to what you need. We don't need any unnecessary conversions.

tarassh:
every account name, contract name, action name, permission name, is converted into uint64, for example `eosio.token -> 0x5530EA033482A600` and `transfer -> 0xCDCD3C2D57000000`. So names are encoded using scheme `[a-z12345.]{1,12}`.

`amount` + `asset` for example, is another structure, which will consist of two fields: amount field + asset filed (asset precision + asset symbol - this will be encoded into uint64)

prusnak:
Ok, use uint64 everywhere it's applicable.

tarassh:
Just to double check.

use `uint64`or `string` for names  in messages?

If we choose `uint64` -> fields will be ready for signing, but anyway we will have to convert it in to text in order to display. 
**The benefit**: we will have more compact message interfaces. **The trade-off**: string data should be converted before sending the message to trezor and again converted into string just to display it.

If we choose `string` -> string field should be verified against `[a-z12345.]{1,12}` regex, and will be ready for displaying and should be converted into `uint64` in order to sign.
**The benefit**: no converting into `uint64` is required before sending message to trezor, data is ready to display. **The trade-off**: regex validation needed, and converting into `uint64` to sign.

matejcik:
Shouldn't the individual `EosAction*` fields be nested within `EosSignTx`? This seems  to be current policy.

`EosActionCommon` is common to all EOS actions (uh). So probably it should be a field in `EosSignTx` instead of repeating across the actions.

I'd maybe keep `from` and `to` as strings, to match what `*GetAddress` returns (and also to match other coins - it makes sense to not pre-parse _addresses_).
It would help to keep the API consistent - OTOH maybe if the addresses actually _are_ `uint64`s then it makes no sense. I'm now trying to find out what EOS docs say about it.

Amounts should definitely be split into fields.

Why do all the actions have `from` or `sender` or whichever? Isn't the source determined through BIP32 derivation of `address_n`?

prusnak:
Use `uint64` for names in messages.

tarassh:
@matejcik 
`EosGetAddress ` I will rename as it is not the base name for this. `EosGetPublicKey` is the proper one.

Now I understand that I need to provide more info about EOS in order for you guys be on the same page with me.

I will start with account **names**. In Eos account names are basically the addresses that are used to send funds. Addresses(Names) are stored in Eos as `uint64` but for end users and any kind of api this `uint64` values are encoded by scheme very close to base32. That was done in order to improve readability and reduce mistakes when you want to send funds to some user `bob.marley` is more readable that long heximal string.

About Keys pairs: 
`5JKQGNiJN6KU7s9fZU74TRG6zraxXTXASMSjaHSyJmA2Tknh3b8` -private key in WIF format
`EOS8ddPoePGrH4x1mha1RHcbjU1cAWWFLWqBQF6Q35RrTsZdLouCm` - public key in WIF format + `EOS` prefix.

Key pairs can be bound to any account, one account could have more that one key pair. Keys are served in order to provide permissions to accounts. For example, **owner** permission is like you have sudo permission to do any operations with your account (change owner key, transfer funds, delegate resources, etc.). **active** permission basically provide the same access level as **owner** except changing owner key. Those two are default, but you can create more if you need.

Now I will explain the transaction structure.
Transaction is composed from actions, actions are atomic operations on EOS blockchain.
actions are basically calling methods in smart contracts. In one transaction you can have actions that are referred to different smart, contracts. 

The structure is something like this:

```
{
     // EOS Transaction Header
    ....
     actions: [
         {
             contract: eosio.token,
             action: transfer,
             permissions: [active],
             data: {
                 from: bob.marley,
                 to: elvis,
                 quantity: '1.0000 EOS',
                 memo: 'are you alive?'
              }
         },
         {
             contract: arbitrary.ico.token,
             action: transfer,
             permissions: [active],
             data: {
                 from: bob.marley,
                 to: elvis,
                 quantity: '1.0000 ICOTOKEN',
                 memo: 'are you alive?'
              }
         },
         {
             contact: eosio,
             action: buyram,
             permissions: [active],
             data: {
                 buyer: bob.marley,
                 receiver: elvis,
                 amount: '1.0000 EOS'
             }
         },

         .......
     ]
}
```

all this and fields  and other not in listed in this json example, are required to sign the transaction.
action json structure can be represented as above or in other format:

```
{
    contract: anycontract,
    action: anyaction,
    permissions: [active],
    data hex: 'long hex string'
}
```

As I understand, its not preferable to have any interface that will accept arbitrary data, that needs to be parsed on trezor side.

So I've created the bunch of actions messages, that should have `ActionCommon` part. as one transaction could have more than 1 action. 

If arbitrary data interface would be accepted, transaction message would look like this.

```
message EosTxSign {
   ...
    required EosHeader header,
    ....
    repeated EosArbitraryAction actions,
    .....
}
```

Regarding EOS documentation. Unfortunately, its scattered across the community resources, there is no single resource, that will answer all questions about EOS. You either have to look for them, or collect knowledge from source code, like I did.

Please let me know if you have any questions.

matejcik:
Thanks for the explanations.

Please switch all `required` fields to `optional`. We don't use `required` to indicate mandatory fields, all non-repeated fields must be `optional`.

If you allow more than one action per transaction, you should not put that into a single message. Instead, Trezor should request each action by `EosTxActionRequest`, and the computer should send the action in a `EosTxActionAck` message. This message would have a `EosActionCommon` field and all `EosAction*` as optional members.
The resulting flow would be:

* **PC**: `EosSignTx { address_n, chain_id, EosTxHeader, num_actions }`
* **Trezor** (`num_actions` times): `EosTxActionRequest`
* **PC** (`num_actions` times): `EosTxActionAck { EosActionCommon, [one of EosAction*] }`
* **Trezor**: `EosSignedTx`

(another option is replying directly with the appropriate `EosAction*`, see what Stellar does. Which of the two you choose mostly depends on how you want to handle the messages in Trezor code; each of the top-level messages gets its own handler, while you'd have a common one for `EosTxActionAck`.)

Speaking of which, please rename `EosSignTxResponse` to `EosSignedTx`.

I still don't understand how "assets" are supposed to work. If I understood you correctly, you get a `uint64` for the amount, and another `uint64` as "precision" (what does that mean?) plus symbol? I'm somewhat surprised that you need to specify symbols separately from the contract IDs; nevertheless, if my understanding is correct, please also switch amounts to `uint64`s. Maybe it makes sense to have a `EosAsset` data type and use it where appropriate?

In `EosActionSellRam`, you specify bytes (which means "how many bytes", right?) as signed `int64`. Is that a typo? If yes, please fix. If not, please switch to `sint64`, our code doesn't support `int64`.

One last thing, why is the `authorization` structure repeated and why does it contain account names? (`actor` is an account name, right?)

tarassh:
I really do like your idea how upload transaction onto device.

Regarding your question about asset. Asset is a struct of two fields 
```
struct asset {
     int64_t amount;
     uint64_t symbol;
}
```

`symbol` field encoded with the same idea like names in EOS. But first byte is used for precision (how may digits after coma particular asset type has).
Other bytes are used to store the asset symbols. Encoding scheme is `(A-Z){1, 12}` - 7bits per symbol.
So, symbols in memory will look something like this: 

`[4digits after coma][EOS.....]`, `[8 digits after coma][ICOTOKEN...]`

I will create `EosAsset ` data type for that.

As for `EosSellRam`, its not a typo. https://github.com/EOSIO/eos/blob/master/contracts/eosio.system/eosio.system.hpp#L181

As you can see `buyrambytes` accepts `uint32_t` for number of bytes and `sellram` accepts `int64_t`.
I cannot say what was the reasons to use signed type, because cpp file has negative value check.

```
void system_contract::sellram( account_name account, int64_t bytes ) {
      require_auth( account );
      eosio_assert( bytes > 0, "cannot sell negative byte" );
```

Currently I am following the defined interface.


About authorization.
Its a part of Action structure
https://github.com/EOSIO/eos/blob/master/libraries/chain/include/eosio/chain/action.hpp#L60

and in the same file we have the definition for `permission_level` struct.
https://github.com/EOSIO/eos/blob/master/libraries/chain/include/eosio/chain/action.hpp#L12

every type there that follows pattern `*_name` is basically the `name` type we were talking previously.

`name` type is used everywhere across EOS blockchain: account names, smart contracts, permissions, method names, etc.

matejcik:
It's certainly possible to build the flow that you describe, esp. if you can process the actions in sequence and don't need to keep them in memory.

Re signed int in `SellRam`, that indeed looks weird. Personally I suspect some sort of technical reason to want a signed type there. Could you clarify with EOS upstream? It would be better if we took an unsigned value, if that is intended.

tarassh:
Found it.

https://github.com/EOSIO/eos/blob/master/contracts/eosio.system/delegate_bandwidth.cpp#L175

In `asset` struct `amount` is signed type. By design you can have negative asset values. 

Here, signed type is used to avoid overflows, otherwise invalid asset conversion may occur.
Negative asset may have bad consequence.
`tokens_out = es.convert( asset(bytes,S(0,RAM)), CORE_SYMBOL);`

conversion implementation:
https://github.com/EOSIO/eos/blob/905e7c85714aee4286fa180ce946f15ceb4ce73c/contracts/eosio.system/exchange_state.cpp#L4

matejcik:
Oh, I just noticed that `BuyRamBytes` is using `uint`**32**. Oh dear that's so ugly I don't even believe it.

Okay so uh
How do these messages actually look like when signing? AFAICT this source code actually implements these method on a node, but how does the transaction look on the wire?

tarassh:
If I understand your question correct.
This is json for sellram:

```
"transaction": {
          "expiration": "2018-08-16T15:51:45",
          "ref_block_num": 41445,
          "ref_block_prefix": 1125993517,
          "max_net_usage_words": 0,
          "max_cpu_usage_ms": 0,
          "delay_sec": 0,
          "context_free_actions": [],
          "actions": [
            {
              "account": "eosio",
              "name": "sellram",
              "authorization": [
                {
                  "actor": "cryptofairy1",
                  "permission": "active"
                }
              ],
              "data": {
                "account": "cryptofairy1",
                "bytes": 10
              },
              "hex_data": "10fc7566d15cfd450a00000000000000"
            }
          ],
          "transaction_extensions": []
        }
}
``` 

`hex_data` is basically buffer used for signing 
`10fc7566d15cfd45`  - `account`
`0a00000000000000` - `bytes`.

Now for buyram:

```
"transaction": {
          "expiration": "2018-08-12T20:59:55",
          "ref_block_num": 58707,
          "ref_block_prefix": 620719579,
          "max_net_usage_words": 0,
          "max_cpu_usage_ms": 0,
          "delay_sec": 0,
          "context_free_actions": [],
          "actions": [
            {
              "account": "eosio",
              "name": "buyram",
              "authorization": [
                {
                  "actor": "cryptofairy1",
                  "permission": "active"
                }
              ],
              "data": {
                "payer": "cryptofairy1",
                "receiver": "cryptofairy1",
                "quant": "0.0005 EOS"
              },
              "hex_data": "10fc7566d15cfd4510fc7566d15cfd45050000000000000004454f5300000000"
            }
          ],
          "transaction_extensions": []
        }
      }
```

`hex_data`
`10fc7566d15cfd45` - `payer`
`10fc7566d15cfd45` - `receiver`
`0500000000000000` - `quant` amount
`04454f5300000000` - `quant` symbol

Now it's definitely clear that ** 64 **.

matejcik:
now I'm confused.

your protobuf has `SellRamBytes` (sell given bytes of RAM), `BuyRamBytes` (buy given bytes of RAM), and `BuyRam` (buy _some number_ of bytes, for given amount of assets).

The JSONs show `sellram` (number of bytes) and `buyram` (number of assets). Is there a JSON equivalent of `BuyRamBytes`?

How would you write the signing code? I.e., how does the code look like that consumes this JSON and produces signatures?

tarassh:
Sorry for confusion, sleepless night.

Buyrambytes json:

```
"transaction": {
          "expiration": "2018-08-14T18:08:48",
          "ref_block_num": 45537,
          "ref_block_prefix": 1199987365,
          "max_net_usage_words": 0,
          "max_cpu_usage_ms": 0,
          "delay_sec": 0,
          "context_free_actions": [],
          "actions": [
            {
              "account": "eosio",
              "name": "buyrambytes",
              "authorization": [
                {
                  "actor": "cryptofairy1",
                  "permission": "active"
                }
              ],
              "data": {
                "payer": "cryptofairy1",
                "receiver": "cryptofairy1",
                "bytes": 100
              },
              "hex_data": "10fc7566d15cfd4510fc7566d15cfd4564000000"
            }
          ],
          "transaction_extensions": []
        }
```

hex_data:
`10fc7566d15cfd45` - `payer`
`10fc7566d15cfd45` - `receiver`
`64000000` - `bytes`

tarassh:
Regarding signing:

signing buffer will be like this:

`[chain_id][header][context_free_actions][actions][transaction_extensions][context_free_data]` -> hash -> `sha256 digest` -> sign secp256k1.

as for `context_free_actions ` - encoded as flat buffer like this: `[size][action1]...[actionN]`, but its always 0, so end buffer is `[size]` no actions. `context_free_actions` - is undocumented.

`actions` - encoded as flat buffer like this: `[size][action1]...[actionN]`.

`transaction_extensions` - is vector<byte>, which is always 0, but we have to encode the size at least.

`context_free_data` - should be hashed if context_free_data.size > 0 otherwise its 32 bytes long 0 value buffer.   context_free_data has some relation to context_free_actions.


end buffer will be
`[chain_id][header][1 byte for size of context_free_actions which is 0][actions][same for transaction_extensions as for context_free_actions][sha256 if context_free_data.size > 0 || 0*32]` 


This is basically encoding scheme before signing. It may be very confusing.

matejcik:
so `actions` is, in your examples, the `hex_data`, plus I assume something that identifies the kind of action, right?

So they actually, for real, encode `BuyRamBytes` as `uint32` and `SellRamBytes as something 64bit. Oh dear. That's actually horrible. I didn't want to believe it, I thought it's an oversight in the C++ API.

Anyway, thanks for the research. ISTM there is not much we can do about it, so we'll keep the same confusion that upstream has. Maybe add a comment to the respective fields that yes, we know it's weird but it copies what upstream does and it's weird on purpose?

Otherwise LGTM, do you have anything else @prusnak?

tarassh:
@matejcik I just found some definition for `context free actions`. I need to investigate it more deeply to understand if its needed in message protocol.

matejcik:
Let's not add these fields now.

adrianbye:
Tarassh : Is this trezor implementation still continuing?  It seems like no updates since Sept 6.  Maybe you need help to get this done?

keepkeyjon:
I've started on a port of this for KeepKey here:
https://github.com/keepkey/device-protocol/tree/eos
https://github.com/keepkey/keepkey-firmware/tree/eos
and FWIW I'm willing to help upstream it to the trezor-mcu codebase once I've got more of the details fleshed out.

protob/messages-eos.proto
```diff
@@ -0,0 +1,176 @@
...
+        optional uint64 account = 1;          // Contract name
+        optional uint64 name = 2;             // Action name
+        repeated EosPermissionLevel authorization = 3;
+    }
+
+    /**
+    * Structure representing transfer data structure
+    */
+    message EosActionTransfer {
+        optional uint64 from = 1;             // Asset sender
```
keepkeyjon: Best choose a different name for these `from` members, as that's a keyword in python.

prusnak:
Superseded by https://github.com/trezor/trezor-common/pull/231

PR #203, opened 2018-09-04
xmr: add integrated addresses indicator

ph4r05:
An integrated address is an address which contains payment ID. Some markets / exchanges are using integrated addresses as it is less prone to errors (payment ID is checksummed).

If a user enters the integrated address as a recipient of transaction it is better to display it in the integrated form, otherwise it is difficult to verify it when confirming the transaction.


prusnak:
Thanks!

PR #204, opened 2018-09-04
Add support for cardano derivation scheme 2

ddeath:
Added parameter for choosing derivation scheme in cardano.

The changes are also in trezor-crypto, trezor-core and python-trezor


prusnak:
In https://github.com/trezor/trezor-crypto/pull/179 we agreed we'll use the v2 scheme exclusively. If there is another scheme in the future, we might consider adding these parameters, but not now.

PR #207, opened 2018-09-08
Add Electurm-KOTO

wo01:


prusnak:
@matejcik can you explain why are some electrum links in coins_details
and some in coins_details.override?

matejcik:
`coins_details.json` is a generated file. `override` is source for these links - AFAIK there can't be links in `coins_details` that don't come from `override`.

we originally moved the Electrum links out of `support.json` because they didn't belong there. But perhaps they should actually go into the coin definition, instead of the override?

prusnak:
Ah, thanks for the explanation. Let's keep it how it is now so we don't bloat coin definitions with stuff that is not used in more places of our infrastructure (AFAIK electrum links are only used on trezor.io/coins not anywhere else).

PR #231, opened 2018-11-06
Add Eos support

atkachyshyn:


matejcik:
did you just re-submit PR #197 ? if not, what are the differences?

atkachyshyn:
Yes, it's evolved resubmit, just from different account.

ср, 7 лист. 2018 о 14:45 matejcik <notifications@github.com> пише:

matejcik:
@atkachyshyn please close one of the PRs then, if you control both

Also please remove the "unknown action" field, this should not be supported per https://github.com/trezor/trezor-common/pull/197#issuecomment-416143356

tarassh:
Unknown action is some kind of equivalent for tx payload in etc https://github.com/trezor/trezor-common/blob/master/protob/messages-ethereum.proto#L67

In order to control Eos account: manage permissions, transfer funds, buy/sell/delegate system resources, actions mentioned in PR were implemented, those are part of contracts that are called system contracts. 

As for Unknown, this type of action should work for any dApps smart contracts developed by users/community.

Like payload in ETH, Unknown should do the same.

protob/messages-eos.proto
```diff
@@ -0,0 +1,265 @@
...
+    optional EosActionLinkAuth link_auth = 12;
+    optional EosActionUnlinkAuth unlink_auth = 13;
+    optional EosActionNewAccount new_account = 14;
+    optional EosActionUnknown unknown = 15;
+     /**
+    * Structure representing asset type
+    */
+    message EosAsset {
+        optional sint64 amount = 1;
+        optional uint64 symbol = 2;           // First 8-bits used for precission.
```
matejcik: shouldn't this read "_lowest_" 8 bits?

prusnak: also precision not precission

atkachyshyn: I have fixed comments, And also did a rebase and squash a few minutes ago. Also for other pull requests i have also performed rebase on master and squash for commits.

atkachyshyn: @matejcik Do you accept @tarassh points\explanation regarding Unknown action, that that's actually equivalent for TX payload in Ethereum integration? Probably the name is a bit misleading. as it's actually not unknown action, as it should be known by EOS ecosystem, we just cannot identify this action as a particular one on the device. If the name is confusing, we can rename it to something more meaningful, like EosActionPayload. So what do you think?

protob/messages-eos.proto
```diff
@@ -0,0 +1,265 @@
...
+        optional uint64 payer = 1;
+        optional uint64 receiver = 2;
+        optional uint32 bytes = 3;       // Number of bytes
+    }
+     /** 
+    * Structure representing sell RAM
+    */
+    message EosActionSellRam {
+        optional uint64 account = 1;
+        optional sint64 bytes = 2;       // Number of bytes
```
keepkeyjon: Should bytes be a `uint64` here?

https://github.com/EOSIO/eos/blob/905e7c85714aee4286fa180ce946f15ceb4ce73c/contracts/eosio.system/eosio.system.abi#L150

protob/messages-eos.proto
```diff
@@ -0,0 +1,265 @@
...
+    */
+    message EosPermissionLevel {
+        optional uint64 actor = 1;
+        optional uint64 permission = 2;
+    }
+
+    /**
+    * Structure representing auth key
+    */
+    message EosAuthorizationKey {
```
keepkeyjon: If you add an optional `repeated uint32 address_n`, it allows for better UX. Instead of having to remember/write down pubkeys in cases where the device can compute it, in those cases you can show the BIP32 path instead.

matejcik:
I still dislike the unknown actions.

@keepkeyjon you seem to understand this better. What is stopping me from passing raw transaction bytes (say, "send many assets to my account") as `EosActionUnknown` message? Does the signature include action type?

keepkeyjon:
> Does the signature include action type?

yes, this is committed to in each action's hash. see:
https://github.com/keepkey/keepkey-firmware/blob/8b4e608a14b06b93608a4e60bf82b2e29f790c26/lib/firmware/eos-contracts/eosio.token.c#L94
https://github.com/keepkey/keepkey-firmware/blob/8b4e608a14b06b93608a4e60bf82b2e29f790c26/lib/firmware/eos.c#L294

If the implementation of the hasher for ActionUnknown checks for "known" actions as I've done [here](https://github.com/keepkey/keepkey-firmware/blob/8b4e608a14b06b93608a4e60bf82b2e29f790c26/lib/firmware/eos.c#L348), then an attacker won't be able to use an unknown action to trick a user into transferring assets, since they won't be able to use the `transfer` action on the token contract without going through the flow that shows what's actually happening.

I personally don't love ActionUnknown either, which is why I've stuck it behind an opt-in [chicken bit](https://github.com/keepkey/keepkey-firmware/blob/8b4e608a14b06b93608a4e60bf82b2e29f790c26/lib/firmware/eos.c#L355). Maybe it makes sense to do the same thing on Trezor (through `ApplySettings`?)

protob/messages-eos.proto
tarassh: Can you describe the case when it will be applicable?

keepkeyjon: Suppose you want to `UpdateAuth` and move your `@active` signing permission to a different HD path on your device. The device-side ux could be something like:

    Do you want to assign memememememe@active permission to EOS Account #1 @active?

assuming it was a valid slip48 path, or:

    Do you want to assign memememememe@active permission to m/42'/17'/32'/57?

if it was not. Contrast that with not providing an `address_n`, which forces the dialog to be:

    Do you want to assign memememememe@active permission to EOS62cPUiWnLqbUjiBMxbEU4pm4Hp5X3RGk4KMTadvZNygjX72yHW?

I've implemented that in our port here: https://github.com/keepkey/keepkey-firmware/blob/7a9db57cd4b15304b478443ea2b2ecdb46c51668/lib/firmware/eos-contracts/eosio.system.c#L507

tarassh: On the device side it looks handy. But It may bring inconvenience on client side. Signed Transaction body has to have key in plane format “EOS5kuQ...”, “PUB_K1_5kuQ...” or “PUB_R1_...” so you have to retrieve key from the device for a new path before generation the transaction body.

keepkeyjon: address_n here is additional *optional* data, so it allows better ux in these case but doesn't force it. Clients are free to ignore it in our implementation, and the wif version of the pubkey would be shown in that case.

keepkeyjon: In practice, it's very easy to handle client-side, since when transferring an account within the device, the UI is going to know what the path and pubkey are well before the device is asking for confirmation.... if you got to that point without this, you would have needed the pubkey anyway.

PR #232, opened 2018-11-08
Added BitCash

WillyTheCat:
Add support for BitCash


prusnak:
Fix build failure - there are couple of tests failing.

Another note: this coin will never be accepted into Trezor in its current form as it uses the same address format as Bitcoin. If you want this supported on Trezor, approach the coin developers to change the address version.

WillyTheCat:
I am the developer of BitCash. BitCash indeed uses the 0 as version byte for the addresses. But the address itself is much longer than a Bitcoin address. For example, this is a valid BitCash address:
17YoNDb7fEpNkkgQ8HaxgFVAoMoEnimPxiV5qmmvtLWjKWbkL7B
Is this a problem? 
How can I reproduce the build failure? Do you have any hints for me?

WillyTheCat:
Okay, I made some changes. Now the tests shows no error any longer. I will add the support for the new prefix byte for the addresses to the BitCash wallet.

prusnak:
Remove check.sh file you created. Remove backend URL, this is not a valid bitcore backend.

WillyTheCat:
Okay, I removed the file and the URL now.

PR #244, opened 2018-11-30
Replace network with protocol magic on cardano

refi93:
See https://github.com/trezor/connect/pull/272 and https://github.com/trezor/trezor-core/pull/417


refi93:
@prusnak thanks for pointing out, fixed

prusnak:
@jpochyla @matejcik are  you OK with this change?

matejcik:
there should be a corresponding PR to python-trezor, which is using the field
otherwise LGTM, except that I don't know how Cardano works so I trust @refi93 that this is correct from that angle :)

refi93:
@matejcik there already is one: https://github.com/trezor/python-trezor/pull/342 I just need to fix the clicks emulation so the test does not hang. Will do probably tomorrow.

PR #247, opened 2018-12-07
Fix build with python 3.6

ehaupt:
Fix build with python 3.6

### DIFF
--- a/tools/coin_info.py
+++ b/tools/coin_info.py
@@ -25,7 +25,7 @@ def load_json(*path):
     else:
         filename = os.path.join(DEFS_DIR, *path)
 
-    with open(filename) as f:
+    with open(filename, encoding="utf-8") as f:
         return json.load(f, object_pairs_hook=OrderedDict)
 
 


prusnak:
What does `import locale ; locale.getpreferredencoding()` return? What OS are you on?

ehaupt:
```console
# python3.6
Python 3.6.7 (default, Dec  4 2018, 01:10:52) 
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)] on freebsd11
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
'UTF-8'
```

matejcik:
I'm inclined to say that non-UTF locales are unsupported.

I mean, we *could* explicitly specify encoding everywhere and it's a technically correct solution. OTOH, all our tooling is based on Click, which very explicitly [does not support ASCII locales](https://click.palletsprojects.com/en/7.x/python3/), so it's somewhat pointless.
So explicit encoding would only support other non-UTF locales (iso8859-*? shift-jis?) which are rather rare these days and personally I'm okay with declaring them unsupported.

What do you think @prusnak?

prusnak:
I am fine with declaring non-UTF-8 locales as unsupported.

matejcik:
alright, closing

PR #248, opened 2018-12-11
Add Florincoin support

KsaRedFx:


KsaRedFx:
Looks like a server outage that happened right when you checked. 
Issue is currently being resolved.

KsaRedFx:
Its up

PR #251, opened 2018-12-13
include AXE in the firmware

charlesrocket:


charlesrocket:
😞 Thought `connect` and `webwallet` will switch with the release process, what would be the best way to approach this after pushed tag?

charlesrocket:
@matejcik got it! Ill remove both git tags and switch to 1.7.3

PR #267, opened 2019-02-01
Tezos: add support for staking

adonagy:
This adds new protobuf messages to Tezos for signing staking operations such as endorsmenets and freshly baked block headers.

We are also ready to create PRs to repositories trezor-core and python-trezor


matejcik:
LGTM, given that we want the "enable staking" toggle to work this way?

adonagy:
I modified the protobuf messages based on https://github.com/trezor/trezor-core/pull/473

PR #273, opened 2019-02-09
defs: add stakenet.json and stakenet.png

JonSalazar:
Add a bitcoin-like coin stakenet.json and stakenet.png


defs/bitcoin/stakenet.json
```diff
@@ -0,0 +1,43 @@
...
+      "High": 200
+    },
+    "dust_limit": 1000,
+    "blocktime_seconds": 60,
+    "uri_prefix": "stakenet",
+    "min_address_length": 27,
+    "max_address_length": 47,
+    "bitcore": [],
+    "blockbook": [
+      "https://xsnexplorer.io"
```
prusnak: remove this, xsnexplorer is not a valid blockbook backend

AlexITC: It isn't, it just supports the `/api/<txhash>` that is required by trezorctl for signing transactions.

If you still think we should remove it, we'll do it.

prusnak: I really doubt it implements the same API as our blockbook if it is not the blockbook.

AlexITC: It doesn't, it just has the necessary endpoint for signing transactions with trezorctl.

prusnak: @matejcik what do you think?

AlexITC: For example, call `GET https://xsnexplorer.io/api/tx/36000e057671b7fc37621bb1a8a547a487198b3cb55e4d8f402abc70e6ecae79` and you'll see that it returns the necessary fields for signing a transaction with trezorctl, `version`, `vin` `vout`, `locktime`.

AlexITC: Using those values I got this transaction generated from trezorctl: https://xsnexplorer.io/transactions/1236786df0c03f03ef74061130017646c8d04bb746a978c3e8555e1c0a0026a7

Screenshot:
img

AlexITC: In fact, thinking a bit more, the unknown is whether this would affect other projects, in the case of trezorctl, the behavior is different when the blockbook is there.

Unless you are confident that this won't affect other projects, I think it is better to remove the blockbook.

matejcik: the "blockbook" field is also used in Trezor Wallet, if we enabled this, and it would fail.

given that some changes are expected here - removal of the blockbook functionality from trezorctl, change of the field format for backends - i wouldn't add it right now.

prusnak: @JonSalazar remove the blockbook entry and I'll merge

AlexITC: @matejcik @prusnak I ended up wrapping a small PR for supporting custom explorers like ours: https://github.com/trezor/python-trezor/pull/358

AlexITC: @prusnak the blockbook links are removed now.

defs/bitcoin/stakenet.json
```diff
@@ -0,0 +1,41 @@
+{
+    "coin_name": "Stakenet",
+    "coin_shortcut": "XSN",
+    "coin_label": "Stakenet",
+    "website": "https://stakenet.io",
+    "github": "https://github.com/X9Developers/XSN",
+    "maintainer": "<support@stakenet.io>",
```
prusnak: maintainer has to be a real person, who is fully responsible for maintaining this definition factually correct at all times, not just some email alias

AlexITC: it is updated.

PR #280, opened 2019-02-18
xmr: Hardfork 10 upgrades

ph4r05:
Monero Hardfork 10 upgrade

- Bulletproof version 2
- Deterministic output commitment masks (based on amount key). Mask balancing performed on inputs.
- EcdhInfo serialization,  serializes just 8B amount


tsusanka:
Same process as before - let's merge this now and review in https://github.com/trezor/trezor-core/pull/478

PR #281, opened 2019-02-27
Fix SmartCash website mistype and add wallet address

leoreinaux:


prusnak:
Never edit coins_details.json. Just modify https://github.com/trezor/trezor-common/blob/master/defs/wallets.json (defs/bitcoin/smartcash.json has already correct values)/

PR #283, opened 2019-02-27
Dash DIP2 transactions support

gladcow:
Dash introduces special transactions as presented in DIP 2 : https://github.com/dashpay/dips/blob/master/dip-0002.md. This changes the transactions such that some transactions will have a variable size payload after the locktime field. 

To create DIP2 transaction in Trezor we need to to add optional `extra_data_len` field in SignTx message.

### DIFF
--- a/protob/messages-bitcoin.proto
+++ b/protob/messages-bitcoin.proto
@@ -134,6 +134,7 @@ message SignTx {
     optional uint32 version_group_id = 8;               // only for Zcash, nVersionGroupId when overwintered is set
     optional uint32 timestamp = 9;                      // only for Capricoin, transaction timestamp
     optional uint32 branch_id = 10;                     // only for Zcash, BRANCH_ID when overwintered is set
+    optional uint32 extra_data_len = 11 [default=0];	// additional tx payload, only for Dash
 }
 
 /**


prusnak:
This should not belong in the `SignTx` message. Transaction created in Trezor will not have any extra data. You should be able to do what you need using `extra_data` and `extra_data_len` fields used by Zcash.

gladcow:
`extra_data` used by Zcash allows to spent funds from transacton with extra data only (i.e. this transaction is created outside Trezor wallet, only its output belongs to Trezor),  not create this transaction. The only way to create such transaction in Trezor is to modify SignTx (I've implemented this functionality here: https://github.com/gladcow/trezor-core/commits/stxes (trezor-core) and https://github.com/gladcow/trezor-mcu/commits/stxes (trezor-mcu). What is the correct way to add this transactions support without modifying SignTx?

prusnak:
@gladcow why would we want to create a DIP2 transaction? Isn't it enough just to spend it?

gladcow:
This is the current list of transactions with extra_data in Dash: https://github.com/dashpay/dips/blob/master/dip-0002-special-transactions.md. This list will be extended in the future with new transaction types. The critical part of such transactions support was ability to spend special transaction,  you are right (mostly because of the special workaround for new masternode registration that allows to register masternode with external collateral and allows to use Trezor's SignMessage function for verification),  but Blockchain User related transactions for example are not available in the Trezor for now.

prusnak:
Trezor should not sign any data it does not understand (and thus cannot show on the display).

Let's focus on spending these special transactions for now (i.e. using just `extra_data` just in the prev-tx).

Later we can add fields to SignTx so Trezor could present and sign the special TX content.

gladcow:
Handling data in prev-tx is working without additional changes,  Zcash has added all required functionality for this case before, so we can spend special txes already. I've hoped to add full support that allows to create transaction with extra_data.

May be you can change your mind if I say that this extra_data provided by the wallet user and concatenated with standard transaction body without any change (so, Trezor can display it at the screen in hex form and user can verify it)? The only difference is the input signs this concatenated data,  not just standard transaction body.

gladcow:
Also I can add support for any specific transaction type separately,  so Trezor can show all payload data in explicit form (for example,  for masternode registration transaction: masternode keys,  operator reward, masternode server address,  etc.). Is it better?

prusnak:
> Also I can add support for any specific transaction type separately

I agree. Let's focus in fixing the broken stuff now (i.e. fixing unspendable TXs which use extra_data).

We can add the extra functionality (creating TXs which us extra_data) later, if there is significant interest in that.

gladcow:
So,  just to be clear: you don't want me to add some functions to create Dash DIP2 transactions without changes in `SignTx` too (I mean message like `DashSignSpecialTx` similar to `CardanoSignTx` or `EosSignTx`)?

prusnak:
Right. Let's just fix DIP2 TX spending  by using extra_data fields already present in the TxAck message.

UdjinM6:
@prusnak Spending DIP2 txes doesn't require any changes or extra_data and works just fine with current firmware and 3rd-party tools like Dash Electrum and Dash Masternode tool. The broken part for spending atm is wallet.trezor.io which can't parse DIP2 txes hex and this should be fixed by https://github.com/BitGo/bitgo-utxo-lib/pull/25. It's creation/signing of special DIP2 txes (version3, types 1-6) that requires `extra_data`.

prusnak:
@UdjinM6 So you are saying we don't need to do any firmware changes in order to spend DIP2 tx?

UdjinM6:
@prusnak Correct, spending issues are not on the firmware side. There were no format changes for inputs (neither for outpoints nor for signatures), so v1/v2 transactions can be created/signed with no issues. The goal of this PR is to take the next step and add full support for DIP2 txes which would let us not only spend but also create/sign such txes via Trezor.

gladcow:
> I find that difficult to believe because Trezor needs to hash the spent transactions in order to compute their hashes so it can verify the correctness of spent transactions?
All required functionality is implemented here https://github.com/trezor/trezor-core/commit/58e4a5546fc98c6a063ac61afd3b9c4bf47ec09a

prusnak:
@gladcow OK, now I understand.

prusnak:
@gladcow OK, now I understand. Coincidentally it was possible to reuse the field extra_data which was designed for Zcash used.

tsusanka:
I hate to close PRs, but we've decided not to support DIP2 transaction in Dash currently. We have limited scope of what we can add to firmware (and mainly maintain), I hope you understand.

Issue #262, opened 2019-01-14
[req] Allow use of other git services


baff5b:
Hello,

Is it possible to use other Git services, like GitLab, when referencing the location for a coin's node source in a definition file? When initially testing, I saw that it would not validate : 
```
github: Value does not match regex ^https://github.com/.*[^/]$
```
when I used a gitlab link. It is possible to mirror the repository to GitHub, but I wanted to know if the repository can be referenced directly.

matejcik:
that shouldn't be a problem, please send a PR with your gitlab value, I'll update the check

prusnak:
Should work now

PR #284, opened 2019-02-27
Add Bitcoin Rhodium

baff5b:
This adds support for Bitcoin Rhodium (https://www.bitcoinrh.org).

There are a few things:
 1. Source release is scheduled to be in March.
 2. Source will be at GitLab. I mentioned in a ticket about this: https://github.com/trezor/trezor-common/issues/262. The validator will not run correctly otherwise.
 3. There are blockbook servers, but there is no SSL certificate yet for it, which I am going to be resolving soon (and can understand it can be a blocker.) There is a blockbook server up now though on HTTP (http://blockbook1.bitcoinrh.org)


prusnak:
Send PR once the source code is public. We are not adding closed source coins.

PR #287, opened 2019-03-04
Add Qtum support

icodeface:


defs/bitcoin/qtum.json
```diff
@@ -0,0 +1,41 @@
+{
+    "coin_name": "Qtum",
+    "coin_shortcut": "QTUM",
+    "coin_label": "Qtum",
+    "website": "https://qtum.org",
+    "github": "https://github.com/qtumproject/qtum/",
```
matejcik: please strip the trailing `/`, here and in testnet

icodeface: fixed

defs/bitcoin/qtum.json
```diff
@@ -0,0 +1,41 @@
...
+    "minfee_kb": 400000,
+    "signed_message_header": "Qtum Signed Message:\n",
+    "hash_genesis_block": "000075aef83cf2853580f8ae8ce6f8c3096cfa21d98334d6e3f95e5582ed986c",
+    "xprv_magic": 76066276,
+    "xpub_magic": 76067358,
+    "xpub_magic_segwit_p2sh": 77429938,
+    "xpub_magic_segwit_native": 78792518,
+    "bech32_prefix": "qc",
+    "cashaddr_prefix": null,
+    "slip44": 88,
```
matejcik: per [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) this value should be 2301

matejcik:
not sure why you removed the duplicate-overrides entry right now?

defs/bitcoin/qtum.json
icodeface: Hi @matejcik, the 2301 is an incorrect number, see the discussion here https://github.com/satoshilabs/slips/pull/196

icodeface:
> not sure why you removed the duplicate-overrides entry right now?

I just saw 

> External contributors should not make changes to `duplicity_overrides.json`, unless
> asked to.

in the https://github.com/trezor/trezor-common/blob/master/defs/README.md

defs/bitcoin/qtum.json
matejcik: could be, but per the same discussion and https://github.com/satoshilabs/slips/pull/120 number 88 is already taken and we don't allow duplicates in our codebases.

matejcik:
well spotted, but in this case I'd be making the exact same change, so please go ahead and put it back there :)

defs/bitcoin/qtum.json
icodeface: changed to 2301

matejcik:
merged, thanks

PR #292, opened 2019-03-09
protob: Add SignUtxo

JonSalazar:
This adds a message that is required for creating TPoS contracts (the Stakenet's way to stake coins).

### DIFF
--- a/protob/messages-bitcoin.proto
+++ b/protob/messages-bitcoin.proto
@@ -137,6 +137,18 @@ message SignTx {
 }
 
 /**
+ * Request: Ask device to sign a transaction input
+ * @end
+ */
+message SignTxInput {
+    optional string coin_name = 1 [default='Bitcoin'];                  // Coin to use for sign
+    repeated uint32 address_n = 2;                                      // BIP-32 path to derive the key from master node
+    required bytes prev_hash = 3;                                       // Hash of previous transaction output to sign
+    required uint32 prev_index = 4;                                     // Index of previous output to sign
+    optional InputScriptType script_type = 5 [default=SPENDADDRESS];    // Defines template of input script
+}
+  
+/**
  * Response: Device asks for information for signing transaction or returns the last result
  * If request_index is set, device awaits TxAck message (with fields filled in according to request_type)
  * If signature_index is set, 'signature' contains signed input of signature_index's input
--- a/protob/messages-bitcoin.proto
+++ b/protob/messages-bitcoin.proto
@@ -137,17 +137,23 @@ message SignTx {
 }
 
 /**
- * Request: Ask device to sign a transaction input
+ * Request: Ask device to sign a utxo
  * @end
  */
-message SignTxInput {
-    optional string coin_name = 1 [default='Bitcoin'];                  // Coin to use for sign
-    repeated uint32 address_n = 2;                                      // BIP-32 path to derive the key from master node
-    required bytes prev_hash = 3;                                       // Hash of previous transaction output to sign
-    required uint32 prev_index = 4;                                     // Index of previous output to sign
-    optional InputScriptType script_type = 5 [default=SPENDADDRESS];    // Defines template of input script
+message SignUtxo {
+    optional string coin_name = 1 [default='Bitcoin'];  // Coin to use for sign
+    required TxInputType inputs = 2;                    // Utxo to sign
 }
-  
+
+/**
+ * Response: Signed utxo
+ * @end
+ */
+ message SignedUtxo {
+    optional string address = 1;    // address used to sign the utxo
+    optional bytes signature = 2;   // signature of the utxo
+}
+
 /**
  * Response: Device asks for information for signing transaction or returns the last result
  * If request_index is set, device awaits TxAck message (with fields filled in according to request_type)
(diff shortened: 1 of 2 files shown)


prusnak:
This is not acceptable. Trezor will not sign individual inputs. Also we are not signing arbitrary input bytes.

If you need something added, modify the `TxInputType` object. Never add arbitrary bytes. You only add value fields and Trezor will serialize the input for you.

AlexITC:
Sorry, I don't get what you ask for.

In order to enable staking, we need to be able to sign a `COutPoint`. The reason is that we have a special transaction for creating what we call `TPoS contract`, such contracts need the following requirements:
2. The transaction spend coins from legacy addresses.
1. The transaction sends 1 coin to the staking address (which must be a legact address).
3. Additionally, an extra output (`OP_RETURN`) is included, it has the contract details:
3.1. The hex-encoded staking address.
3.2. The hex-encoded merchant address.
3.3. The percentage of the rewards going to the staker.
3.4. A signature that is generated with the staking address by hashing the first `OutPoint`.

You can see a TPoS contract on our [explorer](https://xsnexplorer.io/transactions/2f3f49ef22679608d7759c69e210422ae845ca590ae17faa3836d544eca15440), (output 2 has the contract details).

The contract is giving permission to the merchant to stake the coins that the staking address holds, every block reward is split between the merchant and the staker based on the contract reward percentage, you can see a block generated by a contract [here](https://xsnexplorer.io/blocks/18f57bda087d74562859fde575f7571d81fd9a740a04f48b831aed4bab0627d6).

The simpler way that we saw to get this integrated, is by adding a new message to let trezor sign the actual OutPoint, trezor currently displays the OutPoint while asking the trezor owner to confirm if he wants to sign it.

The `TxInputType` seems to be enough to hold the OutPoint but I don't see how we can use it without adding another message or adding the redundant coin field to it, in any case, this will lead to a method missing a meaningful name because the methods name need to match the message name.

Please let us know what could be done.

Thanks.

matejcik:
If i understood your problem earlier, you are trying to sign bytes of an individual input, not a whole transaction. Which should be safe.
You should be able to use the following message:

```protobuf
message SignTxInput {
	optional string coin_name = 1;
	optional TxInputType input = 2;
}
```

`TxInputType` already has a field for `address_n`, script type and prevout, which is all you need.

The problem here is that you are sending the output as undistinguished bytes. You could also use separate fields, like `bytes prev_tx_hash` and `uint32 prev_index`

AlexITC:
We'll get that updated soon, thanks!

JonSalazar:
@matejcik done, please let me know if you need anything else.

The `TxInputType` model wasn't used because it is nested inside another, if we want to use it, we need to extract it or created a nested one on `SignTxInput`.

matejcik:
how would you present confirmation to the user that they're signing what they want?

Would you just display the tx hex / vout? Is that good enough? Perhaps Trezor should validate that the outpoint matches the contract, so you'd need to send the previous tx?

prusnak:
Let's use the following:

```
message SignTxInput {
	optional string coin_name = 1;
	optional TxInputType input = 2;
}
```

Also isn't `SignUtxo` better name?

AlexITC:
@prusnak as I pointed out in a previous comment, the problem with using `TxInputType` is that it is defined inside `TransactionType`, which is defined inside `TxAck`. This means, that we would need to move `TxInputType` to the same level as `SignTxInput`. Here I'm not sure if this breaks any compatibility.

About the name, `SignUtxo` is better.

@matejcik 
Currently, trezor just shows the `[txid]:[index]` text (like `8e3cbb433a7f6a8aebc1efaabb7e6fc96e82837cf1b060eeb28ff226e95ff46d:0`). I believe it is not good enough but haven't found a better.

> Perhaps Trezor should validate that the outpoint matches the contract, so you'd need to send the previous tx?

The utxo that is equal to the first input of the transaction (which is signed by the staking address), having the previous tx doesn't help.

What I see that can actually help is to display the addresses involved in the contract, as well as the commission, in this case, I'm not sure whether that can be done easily because trezor displays the `OP_RETURN` string which we would need to parse to display.

Thanks.

prusnak:
> Here I'm not sure if this breaks any compatibility.

This is not a problem.

> About the name, SignUtxo is better.

Ok, let's use this name then.

What message should be returned from the device? It seems you are missing one. Maybe defining a new message `SignedUtxo`?

AlexITC:
Currently, it returns `MessageSignature`.

prusnak:
Please create a new message:

```
message SignedUtxo {
    optional bytes signature = 1;
}
```

matejcik:
> Currently, trezor just shows the `[txid]:[index]` text (like `8e3cbb433a7f6a8aebc1efaabb7e6fc96e82837cf1b060eeb28ff226e95ff46d:0`). I believe it is not good enough but haven't found a better.

well, how can the user know that they are signing the right utxo? I mean, they can check what the computer tells them, but how do they know the computer is telling them the right thing?

AlexITC:
@prusnak we'll add that, I think is worth returning the address that signed too.

@matejcik 
> well, how can the user know that they are signing the right utxo? I mean, they can check what the computer tells them, but how do they know the computer is telling them the right thing?

That's something difficult, gladly, the user can't loss any coins if the user signs the wrong utxo, because this signature is required for the contract only, he won't be able to stake nor he will loss anything.

matejcik:
Closing this PR as part of cleanup. We are migrating to a monorepo at https://github.com/trezor/trezor-firmware.
Please apply these changes on top of the monorepo and add an implementation in `core`, support for trezorlib and tests in `python`, then send a PR against the monorepo.

Also an implementation note: ISTM we'll need to ask the PC for the previous transaction, to validate the txhash/prevout pair. Otherwise this would allow you to sign 36 arbitrary bytes (prevhash+prevout), we don't want that.
Actually, please tell us what is the exact data that you're signing.

AlexITC:
@matejcik the exact data is the pervious txid and its output index.

So, you want a single PR for everything there?

matejcik:
yes please.
Also as I said, we'll need to verify the previous tx, so that the user is not signing arbitrary bytes (unless you could modify the stake data to e.g. add a constant prefix, like SignMessage does). This should probably be modeled after Bitcoin's TxRequest/TxAck flow for verifying previous txes.

You could also use this data to confirm the staked amount, if that's useful?

AlexITC:
The staked amount is whatever amount the staking address holds, currently there is no support for a prefix but I'll check if it is something we can do.

Thanks.

PR #302, opened 2019-03-26
Enable XSN support on connect

JonSalazar:
Required setting to support Stakenet on [connect](https://github.com/trezor/connect)

### DIFF
--- a/defs/support.json
+++ b/defs/support.json
@@ -11,6 +11,7 @@
       "bitcoin:LTC": true,
       "bitcoin:NMC": true,
       "bitcoin:TEST": true,
+      "bitcoin:XSN": true,
       "bitcoin:VTC": true,
       "bitcoin:ZEC": true,
       "eth:AKA": true,


prusnak:
You don't have any backends defined in https://github.com/trezor/trezor-common/blob/master/defs/bitcoin/stakenet.json

Please add them. Once these are added, deploy Connect locally and try whether everything works:
SignTransaction, PushTransaction, ComposeTransaction, etc.

AlexITC:
We are using our own backend (https://xsnexplorer.io), from what I understand, we can call `signTransaction` passing all the arguments which avoids requiring any backend on the json.

Thanks.

prusnak:
We will merge the PR once this is fixed: https://github.com/trezor/connect/issues/369

prusnak:
Added in https://github.com/trezor/connect/commit/d3a9e8508ab74d05087b84c1b4ebb8a9fa265fc7

AlexITC:
Thanks!

JonSalazar:
You're right, the change is unnecessary, thanks.

PR #308, opened 2019-04-05
Polis Support

eabz:
Hello. 
I will like to add support for Polis. We don't have a trezor capable wallet right now, but we are working on the integration directly to our multi-currency wallet.

PR for Blockbook is being worked: https://github.com/trezor/blockbook/pull/118

If anything else is missing, please let me know.


prusnak:
Squash the commit please

PR #309, opened 2019-04-08
Add Onegram support

Adman:


prusnak:
Let's not add this insignificant coin. Feel free to create a fork of the firmware and deliver it to Onegram users.

PR #315, opened 2019-04-18
mod fujicoin tx fees

fujicoin:
The following changes will be made to the minimum remittance fee from now on as Fujicoin was attacked by Spam Tx.
Before change: 0.001 FJC / kB
After change: 0.1 FJC / kB

Currently, the pool equivalent to about 95% of the network hash rate has completed measures. These pools do not handle low fee remittances. Therefore, the low fee remittance takes about 20 blocks (20 minutes) on average.

Please change Fujicoin remittance fee.

Thanks.


prusnak:
Please send a new pull request to the monorepo: https://github.com/trezor/trezor-firmware

PR #318, opened 2019-04-19
Create Lucent.png

C1ph3r117:
main image logo for main-net


matejcik:
please submit to trezor-firmware repository instead

also the standard process is you create a single PR with all the changes combined

PR #319, opened 2019-04-19
Create lucent_testnet.png

C1ph3r117:
image for testnet lucent


matejcik:
please submit to trezor-firmware repository instead

also the standard process is you create a single PR with all the changes combined

PR #328, opened 2019-07-07
common/defs: add Particl & testnet support

kewde:
I saw that this repo is being kept in line with the monorepo still, or is it preferred there and then backported here?


kewde:
Nevermind, I see the closed PRs. It's being backported here, np!

PR #336, opened 2023-01-04
chore: change cardano testnet

vladimirvolek:
https://github.com/trezor/trezor-suite/pull/7262


matejcik:
this is a read only repo, submit the PR to trezor-firmware instead

PR #338, opened 2023-11-16
Swap `max_gas_fee` and `max_priority_fee` in eip1559 fixtures

marekrjpolak:
When running newly added EIP-1559 serialization in Connect e2e tests, I'm getting `maxFeePerGas cannot be less than maxPriorityFeePerGas (The total must be the larger of the two)`.

I assume it's a general rule for these values so I swapped them in `trezor-common` fixtures and regenerated the signatures.


matejcik (approved):

matejcik:
this repository is read only, please resubmit to `trezor-firmware`

